PR #70, opened 2014-02-26
Use DATABASE_URL only instead of database.yaml

discordianfish:
This makes the Dockerfile set DATABASE_URL instead of mixing it with
database.yaml which apparently doesn't work.
We choose to use DATABASE_URL over database.yaml because it's much
easier to configure/customize this in docker and similiar environments.

You can create the database by:

```
$ docker run -t -i -link mysqld:db \
   DATABASE_URL="mysql2://<user>:<pass>@[HOST]/<db>" \
   rake db:setup
```

(You can also set DATABASE_URL if you can use docker links)

Now you can run promdash like this:

```
$ docker run -t -i -link mysqld:db \
   DATABASE_URL="mysql2://<user>:<pass>@[HOST]/<db>" \
   -p 3000 promdash
```

[HOST] gets replaced by ip:host of the linked container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ ADD        . /promdash
 RUN        bundle install --without="development test migration" && \
            bundle exec rake assets:precompile
 RUN        printf '#!/bin/sh\n \
-           [ -z "$DATABASE_URL" ] && \
-           export DATABASE_URL=$(echo $DB_PORT_3306_TCP|sed 's/^tcp/mysql2/')/$PROMDASH_MYSQL_DATABASE\nexec $@' > \
-           run && chmod a+x run
+           [ -z "$DATABASE_URL" ] && { echo "DATABASE_URL not set"; exit 1; }\n \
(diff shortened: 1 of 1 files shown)


stapelberg:
The commands posted in the original post don’t actually work (they are missing the -e flag for the DATABASE_URL, the first one is missing the container name and uses “rake” instead of “rake2.0”). This is what worked for me:

```
docker run --name prom-mysql -e MYSQL_ROOT_PASSWORD=foobar mysql:5.5

docker run --link prom-mysql:db -e DATABASE_URL="mysql2://root:foobar@[HOST]/prom" \
  prom/promdash /usr/bin/rake2.0 db:setup

docker run --link prom-mysql:db -e DATABASE_URL="mysql2://root:foobar@[HOST]/prom" \
  -p 127.0.0.1:3000:3000 prom/promdash
```

I think you should also put that in the README.md :).

juliusv:
Thanks! Paging @discordianfish!

Issue #71, opened 2014-03-03
First graph is higher in fullscreen view


juliusv:
Steps to reproduce:

1. Create dashboard with at least two graphs
2. Enable fullscreen mode
3. The first graph is multiple pixels taller than the others and the layout of the other graphs around it is changed as a result:

![higher_first_graph](https://f.cloud.github.com/assets/538008/2309150/3e0687bc-a2cc-11e3-8b78-592d234fbccc.png)

grobie:
Can't reproduce. Can this and #178 be close? @juliusv @stuartnelson3

stuartnelson3:
:+1:

juliusv:
Indeed, haven't seen this problem in a year or so. Closing.

PR #178, opened 2014-06-18
Fix mis-aligning fullscreen widgets

stuartnelson3:
Drop the hammer.

image

addresses #71


stuartnelson3:
closed due to staleness

Issue #114, opened 2014-04-04
visual to display point-in-time prometheus results


tomcz:
Current graphs display time-series data from prometheus, but prometheus is also able to provide single value results (eg. number of open/closed circuit-breakers). It would be good to be able to visualize these as a pie chart/colour chart/etc.

PR #189, opened 2014-09-22
Add pie charts to promdash

stuartnelson3:
See title. Plz comment with anything I've missed, or additional functionality you would like like.

requested in #114 

@tomcz @juliusv @grobie @ohookins

### DIFF
--- a/app/assets/javascripts/angular/controllers/frame_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/frame_ctrl.js
@@ -5,7 +5,6 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
                                                     "InputHighlighter",
                                                     "WidgetLinkHelper",
                                                     "GraphiteTimeConverter",
-                                                    "ModalService",
                                                     "CheckWidgetMenuAlignment",
                                                     function($scope, $sce,
                                                              $timeout,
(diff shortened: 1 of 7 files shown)


trompx:
Hello,

Is there a way to display % and the stat value (like the haproxy_server_http_responses_total) in a promdash pie chart ?
Something like http://stackoverflow.com/questions/28306308/how-to-draw-labels-on-dimple-js-donut-or-pie-chart/28342150#28342150 

If not on the graph is there a way to put this info in the legend ?

trompx:
Other problem, on latest version of chrome (I haven't tested on other browsers), when I set the 2 or 3 columns layout (with any aspect ratio), when I hover the pie chart, tooltip is not showing correctly for some parts.

pie chart overflow

When the size of the chart is bigger all is fine though.

stuartnelson3:
>Is there a way to display % and the stat value (like the haproxy_server_http_responses_total) in a promdash pie chart ?
Something like http://stackoverflow.com/questions/28306308/how-to-draw-labels-on-dimple-js-donut-or-pie-chart/28342150#28342150

hey, currently there is no way to edit the pop-up display. if percentage is something you're interested in, it could be added [here](https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/pie_chart.js#L52-L56) as an additional value.

>If not on the graph is there a way to put this info in the legend ?

the legend information can be modified in the legend tab on the pie chart:

image

trompx:
Thanks for pointing to the additional value @stuartnelson3.

As for the legend tab, I already use it. I meant is there a way to get the value (not one of the label like "job", "instance" etc.) but for instance display "72" in the legend for haproxy_frontend_connections_total stat.

stuartnelson3:
unless it's a value being returned from prometheus, no

Issue #101, opened 2014-03-31
Autodect graphite links


grobie:
During frame creation, automatically set the graphite flag if the URL looks like a graphite URL.

Also, show the option to define a frame as graphite frame already in the create screen.

PR #223, opened 2014-10-31
Auto-detect graphite widget links.

stuartnelson3:
Requested in #101


stuartnelson3:
@grobie @juliusv what's the word on this PR? Would be nice to close or merge.

grobie:
I didn't mean it to match on the domain name, but on the path (graphite URLs are always on `/(.../)render`) and parameter names.

PR #257, opened 2014-12-24
Add resolution slider to dashboard.

stuartnelson3:
slider exists both globally, and can also be set for individual graphs.

changing the global resolution reloads all the graphs. changing the individual graph resolution does not reload that graph. this is consistent with the behavior when changing global/graph time ranges. let me know if you want this to behave differently.

### DIFF
--- a/app/assets/templates/graph_template.html.erb
+++ b/app/assets/templates/graph_template.html.erb
@@ -51,16 +51,32 @@
           <label>Graph Title:</label>
           <input class="form-control" ng-model="graph.title">
           <br/>
-          <div class="col-xs-12">
-            <label>Interpolation:</label>
-            <div class="radio">
-              <label class="graph_settings_image cardinal"><input type="radio" ng-model="graph.interpolationMethod" value="cardinal"> cardinal</label>
+          <div class="row">
+            <div class="col-xs-6">
+              <label>Interpolation:</label>
+              <div class="radio">
+                <label class="graph_settings_image cardinal"><input type="radio" ng-model="graph.interpolationMethod" value="cardinal"> cardinal</label>
+              </div>
+              <div class="radio">
+                <label class="graph_settings_image linear"><input type="radio" ng-model="graph.interpolationMethod" value="linear"> linear</label>
+              </div>
+              <div class="radio">
+                <label class="graph_settings_image step"><input type="radio" ng-model="graph.interpolationMethod" value="step-after"> step</label>
+              </div>
             </div>
-            <div class="radio">
-              <label class="graph_settings_image linear"><input type="radio" ng-model="graph.interpolationMethod" value="linear"> linear</label>
-            </div>
-            <div class="radio">
-              <label class="graph_settings_image step"><input type="radio" ng-model="graph.interpolationMethod" value="step-after"> step</label>
+            <div class="col-xs-6">
+              <label>Resolution:
+                <div class="resolution">
+                  <div ui-slider="" min="1" max="5" ng-model="graph.resolution"
+                    class="ui-slider ui-slider-horizontal ui-widget ui-widget-content ui-corner-all" aria-disabled="false">
+                    <a class="ui-slider-handle ui-state-default ui-corner-all" href="#" style="left: 0%;"></a>
+                  </div>
+                  <div>
+                    <span class="pull-left"><i class="icon-minus"></i></span>
+                    <span class="pull-right"><i class="icon-plus"></i></span>
+                  </div>
+                </div>
+              </label>
             </div>
           </div>
 
--- a/vendor/assets/javascripts/ui-slider.js
+++ b/vendor/assets/javascripts/ui-slider.js
@@ -1,0 +1,139 @@
+/*
+ jQuery UI Slider plugin wrapper
+ https://github.com/angular-ui/ui-slider
+*/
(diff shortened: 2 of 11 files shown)


stuartnelson3:
@juliusv

vendor/assets/javascripts/ui-slider.js
```diff
@@ -0,0 +1,138 @@
...
+                    // When ngModel is assigned an array of values then range is expected to be true.
+                    // Warn user and change range to true else an error occurs when trying to drag handle
+                    if (angular.isArray(ngModel.$viewValue) && options.range !== true) {
+                        console.warn('Change your range option of ui-slider. When assigning ngModel an array of values then the range option should be set to true.');
+                        options.range = true;
+                    }
+
+                    // Ensure the convenience properties are passed as options if they're defined
+                    // This avoids init ordering issues where the slider's initial state (eg handle
+                    // position) is calculated using widget defaults
```
juliusv: Here and elsewhere: please put periods at the end of comments, especially in sentence-style comments.

stuartnelson3: That's an external file that I didn't write

app/assets/templates/graph_template.html.erb
```diff
@@ -51,16 +51,33 @@
...
-              <label class="graph_settings_image step"><input type="radio" ng-model="graph.interpolationMethod" value="step-after"> step</label>
+            <div class="col-xs-6">
+              <label>Resolution:
+                <div class="resolution">
+                  <div ui-slider="" min="1" max="3" ng-model="graph.resolution"
+                    class="ui-slider ui-slider-horizontal ui-widget ui-widget-content ui-corner-all" aria-disabled="false">
+                    <a class="ui-slider-handle ui-state-default ui-corner-all" href="#" style="left: 0%;"></a>
+                  </div>
+                  <ul class="resolution_setting_description">
+                    <li>low</li>
```
juliusv: Could we have 5 states? Maybe lowest, low, medium, high, highest.

I'd suggest making "high" as similar as possible to today's default resolution (since today's default resolution is pretty high), and defaulting to high resolution. As for calibration of the high and low ends of the scale, I think the current ends are already pretty good, but could be made a tiny tad more extreme (lowest would be a tiny bit lower than your current low setting, highest would be a tiny bit higher than your current high setting - everything else would be spread evenly between those extremes).

vendor/assets/javascripts/ui-slider.js
juliusv: Ahhh, I didn't notice that without reading the file's path. Could you add something in the header comment about where this is taken from?

juliusv:
This is pretty awesome.

The last thing I'm wondering is whether the slider for the individual graphs should really be in the "Graph and Axis Settings" vs. the "Time Options" tab. Did you have a particular opinion about putting them in the latter. I guess I don't really prefer one over the other strongly, but first I expected them to be under the time options.

stuartnelson3:
Hmm. My first instinct is actually to look for them under "graph and axis settings", rather than "time options". I think of range for time options, and not necessarily anything to do with the density of datapoints/step size. We can crowd source this to @grobie and @matthiasr

juliusv:
@stuartnelson3 Sounds good to me at least.

matthiasr:
To me time options make more sense; from my perspective range and density
are more closely related. Purely subjective though.

app/assets/templates/graph_template.html.erb
stuartnelson3: k, all set. I'm having issues showing "lowest, low, medium, high, highest". visually there really isn't room. the 5 different options exist on the actual slider, though. is that good? or do you want some type of gradation in the slider, or try to fit the words in?

juliusv: Yep, that's fine! Did you push any changes yet though?

stuartnelson3: pushed just now

juliusv:
On Linux (both Chrome/FF), the word "high" wraps around in the global settings and ends up being displayed in the wrong position under the word "low". Adjust the spacing a bit to make it stay where it's supposed to be...

juliusv:
And can we set the default to "high"? (it's still "medium")

matthiasr:
Maybe just leave the scale abstract? I don't think words like "high" or
"medium" really mean much anyway.

juliusv:
Well, as long as the user knows which direction means what :)

matthiasr:
Maybe some simple coarse/fine symbols? There must be something Unicode can
do for us.

stuartnelson3:
i'm drawing a blank on abstract symbols

stuartnelson3:
@juliusv let me know if i fixed the word wrap... browserstack doesn't have ubuntu support

matthiasr:
In terms of symbols, I'm thinking e.g. [&#x2022;](http://www.fileformat.info/info/unicode/char/2022/index.htm) ––– [&#x22ef;](http://www.fileformat.info/info/unicode/char/22ef/index.htm) or [&#x254c;](http://www.fileformat.info/info/unicode/char/254c/index.htm) ––– [&#x2508;](http://www.fileformat.info/info/unicode/char/2508/index.htm).

stuartnelson3:
awesome, pushed

stuartnelson3:
@juliusv when you get a chance can you show me how you were able to load a page without the resolution selection being respected? can't seem to reproduce.

juliusv:
:+1:

PR #262, opened 2015-01-05
Remove code forcing graph max.

### LINKED ISSUE #258: Redraw graph when selecting/deselecting timeseries
Scenario:
3 timeseries: a with values ~220k, b and c with values ~1000-1500

Current:
Deselecting the label (clicking on the tick in the legend) only hides a timeseries, without redrawing the graph based on the new values / changing the axis.

Expected:
Redraw the graph, so that timeseries with small values become visible.

Related:
* grafana: expected behavior
* newrelic: expected behavior

stuartnelson3:
fixes #258

### DIFF
--- a/app/assets/javascripts/angular/directives/graph_chart.js
+++ b/app/assets/javascripts/angular/directives/graph_chart.js
@@ -224,8 +224,6 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
 
         if (scaleID) {
           rsGraph.max = yScales[scaleID](graphMax);
-        } else {
-          rsGraph.max = rsGraph.renderer.domain().y[1];
         }
 
         var $legend = $el.find(".legend");


stuartnelson3:
@grobie @juliusv

grobie:
I have no idea why that line is there and what it does, but everything seems to do what I expect, so .... :+1: ? :D

stuartnelson3:
The line was originally added in pull request https://github.com/prometheus/promdash/pull/179

it's purpose was to fix a previously incorrect understanding of how rickshaw scales graph data for rendering. unfortunately, the fix resulted in hardcoding the graph height. removing the code appears to solve both the problem in the initial PR (stacked graph data not being drawn correctly) and this current problem.

You can verify that the code is working correctly and solves both issues by changing the `else` branch to set `rsGraphMax` to `1`, then looking at a stacked type graph. Or, if you're lazy, come bother me and I'll show you :P

juliusv:
:+1:

Issue #259, opened 2015-01-02
Preserve state of legend selection


grobie:
Steps:
- open dashboard
- activate refresh for a specific interval
- select/deselect a time series in a graph (using the legend menu)

Expected:
- selection doesn't change after graph refresh

Observed:
- graph refreshes, selection state is lost and needs to be redone.

juliusv:
I can imagine this won't be super straightforward to implement, but :+1: in principle.

stuartnelson3:
each time a graph refreshes, a the old graph is being destroyed and a new graph is being drawn in its place; there is no knowledge between graphs of legend state.

how vital is this? we would have to maintain the state separately, which sounds messy and probably error prone.

stuartnelson3:
closed by #263

PR #263, opened 2015-01-05
Maintain the legend state between graph refreshes.

stuartnelson3:
Requested in #259 

Depends on #262

### DIFF
--- a/app/assets/javascripts/angular/directives/graph_chart.js
+++ b/app/assets/javascripts/angular/directives/graph_chart.js
@@ -54,7 +54,14 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         var graphHeight = WidgetHeightCalculator(element[0], scope.aspectRatio);
         $el.css('height', graphHeight);
 
+        var disabledSeries = {};
         if (rsGraph) {
+          // Create a unique name using the series labels to store legend state for newly drawn graph.
+          rsGraph.series.forEach(function(s) {
+            if (s.disabled) {
+              disabledSeries[s.uniqName] = true;
+            }
+          });
           $el.html('<div class="annotation"></div><div class="graph_chart"><div class="legend"></div></div>');
           rsGraph = null;
         }
@@ -221,6 +228,16 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
           interpolation: scope.graphSettings.interpolationMethod,
           series: series
         });
+
+        // Maintain enabled/disabled state between graph updates.
+        if (Object.keys(disabledSeries).length) {
+          for (var i = 0; i < rsGraph.series.length; i++) {
+            var s = rsGraph.series[i];
+            if (disabledSeries[s.uniqName]) {
+              s.disabled = true;
+            }
+          }
+        }
 
         if (scaleID) {
           rsGraph.max = yScales[scaleID](graphMax);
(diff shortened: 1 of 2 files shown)


stuartnelson3:
@grobie @juliusv

grobie:
Great to have this feature in general! What other information does an object in `rsGraph.series` carry around? Maybe there is some more stable identifier which could be use than the order?

I'd say even the name would of the legend could do it. If the legend name changes because someone changes the format, then it might be fine also just fine to lose the disabled state. Best case would be a reference to the timeseries which is independent of the name, but if that's not available it's not necessary to spend too much time on it imho.

stuartnelson3:
K, creating a unique name based on the entire label name.

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -54,7 +54,10 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         var graphHeight = WidgetHeightCalculator(element[0], scope.aspectRatio);
         $el.css('height', graphHeight);
 
+        var seriesState = {};
         if (rsGraph) {
+          // Create a unique name using the series labels to store legend state for newly drawn graph.
+          rsGraph.series.forEach(function(s) { seriesState[renderLabels(s.labels)] = s.disabled });
```
juliusv: Could we optimize for the common case for how much state we need to keep? For example, only save disabled items in this hash, which should normally be 0? You could even make it more efficient (assuming that usually people will select exactly one timeseries if they do select one) by not saving anything if everything is enabled, and only saving the enabled ones otherwise.

Also, I wouldn't base the identification of a timeseries based on `renderLabels`. It doesn't include the metric name, but it does include superfluous stuff like HTML. The best identification would be `metric_name{label1="value1", ...}` or something like that. Basically, the default name we give to a timeseries in the legend if there is no format string configured.

juliusv: Regarding my first paragraph, you might want to go for the first suggestion, as it will keep the code simpler (just save the disabled ones - you don't need to check first if any are disabled).

stuartnelson3:
Alright, ready for a re-re-re-review

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -54,7 +54,14 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         var graphHeight = WidgetHeightCalculator(element[0], scope.aspectRatio);
         $el.css('height', graphHeight);
 
+        var seriesState = {};
```
grobie: rename to `disabledSeries`?

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -54,7 +54,14 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         var graphHeight = WidgetHeightCalculator(element[0], scope.aspectRatio);
         $el.css('height', graphHeight);
 
+        var disabledSeries = {};
```
stuartnelson3: updated name

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -54,7 +54,14 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
...
         if (rsGraph) {
+          // Create a unique name using the series labels to store legend state for newly drawn graph.
+          rsGraph.series.forEach(function(s) {
+            if (s.disabled) {
+              disabledSeries[s.uniqName] = true;
```
stuartnelson3: setting directly to `true`

stuartnelson3:
ping

juliusv:
Lookin' gooooooooooooood. :+1:

Issue #260, opened 2015-01-04
Visualize relationship between until and refresh options


grobie:
In order to make clear that refresh is set to "never" when a fixed "until" date is provided,
Automatically select "never" in the drop down list and deactivate the form element.

stuartnelson3:
closed by #264

PR #264, opened 2015-01-05
Add watch on the 'until' query parameter that disables/enables the

stuartnelson3:
globalRefresh dropdown button.

requested in #260


stuartnelson3:
@grobie

grobie:
Looks good in general. I believe it should be possible now to simplify the refresh timer setup code, as it doesn't seem to be longer necessary to manually watch for URL variables.

PR #265, opened 2015-01-06
Feature spec for legend state

stuartnelson3:
@grobie

### DIFF
--- a/spec/features/graph/legend_spec.rb
+++ b/spec/features/graph/legend_spec.rb
@@ -15,7 +15,17 @@ feature "Graph legend", js: true do
         expect(page).to have_content "prometheus_targetpool_duration_ms"
       end
     end
+
+    it "maintains legend state after update" do
+      expect(all('.rickshaw_legend .line.disabled').count).to eq 0
+      legend_elements = all('.rickshaw_legend .action')
+      legend_elements.first.click
+      expect(all('.rickshaw_legend .line.disabled').count).to eq 1
+      open_tab 'Refresh'
+      expect(all('.rickshaw_legend .line.disabled').count).to eq 1
+    end
   end
+
 
   describe "two legend strings" do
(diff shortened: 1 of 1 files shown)


spec/features/graph/legend_spec.rb
```diff
@@ -15,8 +15,18 @@
         expect(page).to have_content "prometheus_targetpool_duration_ms"
       end
     end
+
+    it "maintains legend state after update" do
+      expect(all('.rickshaw_legend .line.disabled').count).to eq 0
+      legend_elements = all('.rickshaw_legend .action')
+      legend_elements.first.click
+      expect(all('.rickshaw_legend .line.disabled').count).to eq 1
+      open_tab 'Refresh'
```
grobie: How does open_tab work. Does it make sure that all elements have been updated already? Can we ensure that something which _should have_ updated got updated before the line.disabled check?

stuartnelson3: `open_tab` just finds the 'Refresh' on the graph and clicks it. I added in 3sec sleep statements to make sure the update had occurred but then removed them. Would you like me to add in a 100ms sleep between clicking refresh and checking for .line.disabled? Since the "prometheus" server in the feature specs is just a local sinatra server the response is almost immediate when refreshing.

grobie: In my experience we should try to avoid sleeps in tests to ensure a state change. It does not ensure that a change really has happened, slows down tests and exposes the test suite to potential one-off test errors (slow execution on CI servers, et.).

What data does the test server return? Could we implement it in a way that it's for example a monotonically increasing counter? Then we could add a method which waits/ensures that the last data point has been updated before we check the test expectation.

grobie:
:+1: if the test fails without the new legend preserving code.

Issue #269, opened 2015-01-07
Set any graphite parameters if not graphite


grobie:
If a graphite URL gets pasted which does not include `width`, `height` or `from` parameters, promdash currently does not set the parameters at all, leading to not correctly scaled graphs. Set the parameters even if not present, as it's already the case for the `until`.

grobie:
Fixed with #271.

PR #271, opened 2015-01-07
Automatically add width, height, from, and until query params to graphite urls

stuartnelson3:
requested in #269

### DIFF
--- a/app/assets/javascripts/angular/controllers/frame_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/frame_ctrl.js
@@ -38,38 +38,35 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
   function buildFrameURL(url) {
     var parser = document.createElement('a');
     parser.href = url;
-    var queryStringComponents = parser.search.substring(1).split('&');
     if ($scope.frame.graphite) {
-      if (url.indexOf("until") === -1) {
-        queryStringComponents.push("until=now");
-      }
-      queryStringComponents = queryStringComponents.map(function(e) {
-        switch (0) {
-        case e.indexOf('height='):
-          return setDimension(e, $scope.frameHeight().height);
-        case e.indexOf('width='):
-          var width = $scope.frameHeight().height / $scope.aspectRatio;
-          return setDimension(e, width);
-        case e.indexOf('from='):
-          if (!$scope.frame.range) {
-            return e;
-          }
-          return setDimension(e, GraphiteTimeConverter.graphiteFrom($scope.frame.range, $scope.frame.endTime));
-        case e.indexOf('until='):
-          return setDimension(e, GraphiteTimeConverter.graphiteUntil($scope.frame.endTime));
-        }
-        return e;
-      });
-      queryStringComponents.push("bgcolor=%23191919");
+      return parseGraphiteURL(parser);
     }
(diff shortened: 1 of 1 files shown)


stuartnelson3:
@grobie

grobie:
Isn't it the common javascript style to use single quotes? Assuming @juliusv trusts Google's style guide: https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml

app/assets/javascripts/angular/controllers/frame_ctrl.js
```diff
@@ -36,33 +36,37 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
...
           return setDimension(e, $scope.frameHeight().height);
-        case e.indexOf('width='):
+        case e.indexOf("width="):
           var width = $scope.frameHeight().height / $scope.aspectRatio;
           return setDimension(e, width);
-        case e.indexOf('from='):
+        case e.indexOf("from="):
           if (!$scope.frame.range) {
+            // If the frame doesn't specify a time range, use the time range
+            // in the query string.
```
grobie: This is unexpected. If a dashboard is set to "range 3h" then a graphite frame will only show 1h. I'd vote to reduce the number of branches in the code and to always use the frame range (which should default to the dashboard range if not overridden).

app/assets/javascripts/angular/controllers/frame_ctrl.js
```diff
@@ -36,33 +36,37 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
...
+    var queryStringComponents = parser.search.substring(1).split("&");
     if ($scope.frame.graphite) {
-      if (url.indexOf("until") === -1) {
-        queryStringComponents.push("until=now");
-      }
+      ["height=0", "width=0", "from=-1h", "until=now"].forEach(function(e) {
+        // Unshift elements to the front of the query string. If `from`
+        // already exists, it will not be clobbered by these values.
+        queryStringComponents.unshift(e);
+      });
```
grobie: This code becomes more and more complicated, with unclear side effects which require comments to explain what's going on. I might be missing something, but maybe rewriting the whole function including `setDimension` could help here? A first idea:

```javascript
function buildFrameURL(url) {
  var parser = document.createElement('a');
  parser.href = url;
  var queryStringComponents = parser.search.substring(1).split('&');
  if ($scope.frame.graphite) {
    var fields = {};
    queryStringComponents.forEach(function(f) {
      var s = f.split("=");
      fields[s.shift()] = s.join('=');
    });
    
    fields['height'] = $scope.frameHeight().height;
    fields['width'] = $scope.frameHeight().height / $scope.aspectRatio;
    fields['from'] = GraphiteTimeConverter.graphiteFrom($scope.frame.range, $scope.frame.endTime);
    fields['until'] = GraphiteTimeConverter.graphiteUntil($scope.frame.endTime);
    fields['bgcolor'] = '%23191919'

    // join fields
  }

  // ...
}
```

That approach would ensure that the 4 parameters always get set correctly, regardless of what parameters are set already. I also believe that not using this switch statement makes the code more comprehensible.

Ideally, the code to parse/construct an URL should be separated from the business logic to construct a graphite URL.

stuartnelson3:
made changes

grobie:
:+1: but probably depends on #274

PR #273, opened 2015-01-07
Accept a number (integer or float) for resolution value.

### LINKED ISSUE #272: cannot save dashboard at lowest resolution
The json schema is set to only accept integers, but the lowest setting is a float.

stuartnelson3:
fixes #272

### DIFF
--- a/dashboard_schema.json
+++ b/dashboard_schema.json
@@ -32,7 +32,7 @@
           "type": "boolean"
         },
         "resolution": {
-          "type": "integer"
+          "enum": [1, 2, 3, 4, 5]
         },
         "endTime": {
           "type": [
@@ -199,7 +199,7 @@
                 "type": "string"
               },
               "resolution": {
-                "type": "integer"
+                "enum": [1, 2, 3, 4, 5]
               },
               "legendFormatStrings": {
                 "type": "array",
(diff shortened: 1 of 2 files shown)


stuartnelson3:
@juliusv

juliusv:
Hmm... or since the slider only allows setting (and displaying) several concrete values, should it be an enum?

(Ctrl-F for "enum" here: http://json-schema.org/example2.html)

stuartnelson3:
sounds good, pushed.

dashboard_schema.json
```diff
@@ -199,7 +199,7 @@
                 "type": "string"
               },
               "resolution": {
-                "type": "integer"
+                "enum": [0.5, 1, 2, 3, 4, 5]
```
stuartnelson3: `0.5` and `1` are included because the graph resolution value, if set on an individual graph, can be anything `1-5`. It isn't until the graph is refreshed that the resolution will be set to `0.5`

juliusv: Hmm, I had trouble understanding this and found https://github.com/prometheus/promdash/commit/24ac83eca971c99cba70a2afd15d4b81a5a1c367#diff-86fa8341d9271c16a38806d196d801e6R118.

I guess the problem is that you can't tell the slider to directly have the values `0.5`, `2`, ... and so on?

I'd still prefer if that code wouldn't actually change the original slider value though - how about simply making a local transformation to the slider value before using it to compute the `scalingFactor`?

stuartnelson3: +1, changed and pushed

Issue #274, opened 2015-01-08
Always default to dashboard range on widget creation


grobie:
Currently, promdash widgets default to `1h` on creation, graphite widgets default to `undefined`. Let's initialize any new widget with the current dashboard range.

stuartnelson3:
closed by #275

PR #275, opened 2015-01-08
set widget range on creation to dashboard range

stuartnelson3:
requested in #274


juliusv:
Same for `until`?

stuartnelson3:
good point, updated

juliusv:
:+1:

PR #279, opened 2015-01-13
Add `overflow: auto;` to expression form to force line break.

stuartnelson3:
@juliusv 

## before
image

## after
image


juliusv:
:+1:

Issue #261, opened 2015-01-04
Make setting an until date easier


grobie:
Use case:
In order to set "until" to the current time,
I want an option to quickly set date and time to "now",
without having to click through time and date.

In order to clear "until",
I want an option to quickly empty the date and close the form menu,
with a single click.

Idea:
Add two buttons (e.g. above or below the current form) for each action.

grobie:
Maybe just a clear icon at the right end of the input field? 

![](http://i.imgur.com/ulP6mhS.png)

Comments @stuartnelson3 @juliusv

stuartnelson3:
the 'clear' button sounds good to me

setting 'until' to current time: this currently takes *two clicks*. click the input, then click the highlighted current date. do you really need this as one click?

grobie:
@stuartnelson3 I only learned about the short click option later. If the timezone conversion gets fixed, it would indeed solve the described use case.

PR #282, opened 2015-01-19
Allow clearing on hover global and widget end time

stuartnelson3:
requested in #261 

@grobie @juliusv thoughts on hover vs. always visible `x` to clear the input? I like the hover.

### DIFF
--- a/app/assets/templates/widget_time_options_template.html
+++ b/app/assets/templates/widget_time_options_template.html
@@ -26,7 +26,14 @@
             <i class="icon-fb"></i>
           </button>
         </span>
-        <input type="text" class="form-control" datetime-picker datetime="endTime" data-format="yyyy-MM-dd" placeholder="Until">
+        <span class="until_container">
+          <input type="text" class="form-control" datetime-picker datetime="endTime" data-format="yyyy-MM-dd" placeholder="Until">
+          <span class="until_clear pointer"
+            ng-click="clearEndTime()"
+            tooltip-append-to-body="true" tooltip="Clear">
+            <i class="glyphicon glyphicon-remove"></i>
+          </span>
+        </span>
         <span class="input-group-btn">
           <button class="btn btn-default" ng-click="increaseEndTime()" title="Increase end time">
             <i class="icon-ff"></i>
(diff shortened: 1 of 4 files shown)


app/assets/templates/widget_time_options_template.html
```diff
@@ -26,7 +26,14 @@
             <i class="icon-fb"></i>
           </button>
         </span>
-        <input type="text" class="form-control" datetime-picker datetime="endTime" data-format="yyyy-MM-dd" placeholder="Until">
+        <span class="until_container">
+          <input type="text" class="form-control" datetime-picker datetime="endTime" data-format="yyyy-MM-dd" placeholder="Until">
+          <span class="until_clear"
+            ng-click="clearEndTime()"
+            tooltip-append-to-body="true" tooltip="Clear">
+            <i class="glyphicon glyphicon-remove-circle"></i>
```
juliusv: I think `glyphicon-remove` (without circle) would fit better into the adjacent visual language we're using.

Also, give the glyphicon some (6px?) padding/margin to the right so that it has equal distance from the top/bottom of the input field as from the right.

juliusv: Bonus points if the remove icon only appears when the end time is not already set to `null`.

stuartnelson3: skipping the bonus points

app/views/dashboards/show.html.erb
```diff
@@ -26,7 +26,14 @@
       <button class="btn btn-default" ng-click="decreaseEndTime()" title="Decrease end time">
         <i class="icon-fb"></i>
       </button>
-      <input type="text" class="form-control" datetime-picker datetime="globalConfig.endTime" data-format="yyyy-MM-dd" placeholder="Until">
+      <span class="until_container">
+        <input type="text" class="form-control" datetime-picker datetime="globalConfig.endTime" data-format="yyyy-MM-dd" placeholder="Until">
+        <span class="until_clear global pointer"
```
juliusv: The "bonus points" seem quite easy though: simply add this to the span:

    ng-show="globalConfig.endTime"

(same in the widget_time_options_template.html)

grobie: Doubling the bonus points.

stuartnelson3: clever boy

where's my points

juliusv: :cookie: :100:

grobie:
\o/

PR #283, opened 2015-01-19
Remove encoding/decoding of user-facing frame urls.

### LINKED ISSUE #280: Graphite template URL from query string/target pairs should not be escaped
Reproduction:

Show template variables
Add template variable `xxx` to be `packets_received`
Add Frame with URL `http://graphite/render/?target=stats_counts.statsd.packets_received`
Verify graph shows something
Click on Frame Source
In `target` field, replace `packets_received` with template variable `{{xxx}}`

Actual:

Graph no longer shows `No Data`
Frame Source is now:

```
http://graphite/render/?target=stats_counts.statsd.%7B%7Bxxx%7D%7D
```

Expected:

To show the same graph without the template variable
Frame Source to not escape the template braces:

```
http://graphite/render/?target=stats_counts.statsd.{{xxx}}
```

stuartnelson3:
I'm reasonably sure these `encode`/`decode` function calls were added for a reason in the original pull request, but I can't remember why now, and I can't find a comment about them. Do either of you remember what these were for? The frame construction code has since [changed](https://github.com/prometheus/promdash/commit/23897351e2420699a6ddf90538fcef70a343661b), and looking at dashboards with graphite widgets I'm not seeing any breakage, so maybe they aren't needed.

fixes #280 

@juliusv @grobie

### DIFF
--- a/app/assets/javascripts/angular/controllers/frame_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/frame_ctrl.js
@@ -61,10 +61,10 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
     var queryStringComponents = parser.search.substring(1).split("&");
     // [{foo: bar}, {baz: quux}]
     $scope.frameComponents = queryStringComponents.map(function(kv) {
-      var kvArr = kv.split("=");
+      var kvArr = unescape(kv).split(/=(.+)/);
       return {
         key: kvArr[0],
-        value: kvArr[1] === undefined ? undefined : decodeURIComponent(kvArr[1]),
+        value: kvArr[1],
       };
     });
   };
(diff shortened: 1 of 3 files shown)


grobie:
Strong vote to write more tests in this project, as they act as a specification. The rate of "I have no idea why this functionality is here" got way too high recently for my taste.

The encoding and decoding was requested in order to
* properly escape a user entered parameter (e.g. `target` set to `alias(seriesList, "new $&?= name")`)
* make it easier to read/modify a given target by unescaping any special characters

stuartnelson3:
this seem to do the trick?

grobie:
I'm not sure what the behavior is now with this trick. I couldn't find tests nor at least a commit message which describes the reasons for the change in best case with examples. Mixing whitespace changes with behavior changes in one commit also makes it more difficult to see the changes.

Can you please split the commit in two and explain the changes in the commit message? This will also help in the future with questions like "Do either of you remember what these were for?".

stuartnelson3:
split

grobie:
I can't really judge the side effects of only escaping a few characters. In general my experience is that such "fixes" will only postpone the issue and lead to similar issues later. On the other hand, any other solution to handle template variables might require more changes to the code, so it might be a compromise we want to add. @juliusv time ;)

juliusv:
Hey, haven't forgotten about this - it's just queued behind launch-critical
stuff at the moment. Will get to it soon.

juliusv:
This probably needs a rebase on master now?

stuartnelson3:
rebased, seems to still be working

stuartnelson3:
@juliusv ping

stuartnelson3:
@juliusv k, i think this is ready now. I can't think of a good way to support `'&'` in, say, an alias function, but that may just be an acceptable limitation.

stuartnelson3:
@juliusv updated

juliusv:
:+1:

Issue #241, opened 2014-12-01
x-axis zoom


stuartnelson3:
http://code.shutterstock.com/rickshaw/examples/extensions.html

cc/ @jdarc

grobie:
The "drag&zoom" introduction blog post: http://blog.newrelic.com/2011/07/15/drag-and-zoom-for-great-justice/

stuartnelson3:
@grobie @juliusv priority level on this?

juliusv:
I'm fine with this in general, though everything depends on cost of implementation and maintenance. I'll leave that judgement and the prioritization up to @grobie and you :)

I would also prefer the NewRelic way of doing things, even though it doesn't exist as a finished Rickshaw component yet. It feels more natural, and it also requires less UI space. However, one should be careful to get the details right: e.g. how to zoom out (or maybe not necessary in v1)?

grobie:
I have never seen any short cut to zoom out. It'll still be possible to change the date with the existing options, so I don't see zoom out to be part of the scope of this feature. One can also simply reloading the page to reset the default range/until.

grobie:
This is one of the most kick-ass features in promdash. Very well done @stuartnelson3! ^5

![Approved](http://media.giphy.com/media/BQAk13taTaKYw/giphy.gif)

PR #147, opened 2014-04-23
Some keyboard commands

stuartnelson3:
Ok, so this wasn't high on the priority list, but it's cool and it only took 20min to do.

I did hack it out in coffeescript though, so I can always translate it to javascript.

`g d`: go to dashboards (the root)
`g s`: go to the servers
`g b`: go back (hitting your browser back button)
`j`: navigate down a list, servers or dashboards page
`k`: navigate up a lost, servers or dashboards page
`o`: open a highlighted list item
`enter`: open a highlighted list item
`c`: new dashboard/server page (location dependent)
`F`: enter fullscreen on dashboard page

I didn't add `c` support for `dashboard#show` since the graph/clone/frame functionality is all separate.

requested in #147 

@grobie

PR #284, opened 2015-01-20
X axis zoom

stuartnelson3:
requested in #241

### DIFF
--- a/app/assets/javascripts/Rickshaw.Graph.DragZoom.js
+++ b/app/assets/javascripts/Rickshaw.Graph.DragZoom.js
@@ -1,0 +1,122 @@
+Rickshaw.namespace('Rickshaw.Graph.DragZoom');
+
+Rickshaw.Graph.DragZoom = Rickshaw.Class.create({
+
+  initialize: function(args) {
+    if (!args.graph) {
+      throw "Rickshaw.Graph.DragZoom needs a reference to a graph";
+    }
+    var defaults = {
+      opacity: 0.5,
+      fill: 'steelblue',
+      minimumTimeSelection: 60,
+      callback: function() {}
+    };
+
+    this.graph = args.graph;
+    this.svg = d3.select(this.graph.element).select("svg");
+    this.svgWidth = parseInt(this.svg.attr("width"), 10);
+    this.opacity = args.opacity || defaults.opacity;
+    this.fill = args.fill || defaults.fill;
+    this.minimumTimeSelection = args.minimumTimeSelection || defaults.minimumTimeSelection;
+    this.callback = args.callback || defaults.callback;
+
+    this.registerMouseEvents();
+  },
+
+  registerMouseEvents: function() {
+    var self = this;
+    var ESCAPE_KEYCODE = 27;
+    var rectangle;
+
+    var drag = {
+      startDt: null,
+      stopDt: null,
+      startPX: null,
+      stopPX: null
+    };
+
+    this.svg.on("mousedown", onMousedown);
+    this.svg.on("touchstart", onMousedown);
+
+    function pointAsDate(e) {
+      return Math.floor(self.graph.x.invert(e.offsetX));
+    }
+
+    function onMouseup(datum, index) {
+      drag.stopDt = pointAsDate(d3.event);
+      var windowAfterDrag = [
+        drag.startDt,
+        drag.stopDt
+      ].sort(compareNumbers);
+
+      self.graph.window.xMin = windowAfterDrag[0];
+      self.graph.window.xMax = windowAfterDrag[1];
+
+      var endTime = self.graph.window.xMax;
+      var range = self.graph.window.xMax - self.graph.window.xMin;
+
+      reset(this);
+
+      if (range < self.minimumTimeSelection || isNaN(range)) {
+        return;
+      }
+      self.graph.update();
+      self.callback({range: range, endTime: endTime});
+    }
+
+    function reset(el) {
+      var s = d3.select(el)
+      s.on("mousemove", null);
+      d3.select(document).on("mouseup", null);
+      s.on("touchmove", null);
+      d3.select(document).on("touchend", null);
+      d3.select(document).on("touchcancel", null);
+      drag = {};
+      rectangle.remove();
+    }
+
+    function compareNumbers(a, b) {
+      return a - b;
+    }
+
+    function onMousemove() {
+      var offset = drag.stopPX = d3.event.offsetX;
+      if (offset > (self.svgWidth - 1) || offset < 1) {
+        return;
+      }
+
+      var limits = [drag.startPX, d3.event.offsetX].sort(compareNumbers);
+      var selectionWidth = limits[1]-limits[0];
+      if (isNaN(selectionWidth)) {
+        return reset(this);
+      }
+      rectangle.attr("fill", self.fill)
+               .attr("x", limits[0])
+               .attr("width", selectionWidth);
+    }
+
+    function onMousedown() {
+      var el = d3.select(this);
+      rectangle = el.append("rect")
+                    .style("opacity", self.opacity)
+                    .attr("y", 0)
+                    .attr("height", "100%");
+
+      drag.target = d3.event.target;
+      drag.startDt = pointAsDate(d3.event);
+      drag.startPX = d3.event.offsetX;
+      d3.event.preventDefault ? d3.event.preventDefault() : d3.event.returnValue = false;
+      el.on("mousemove", onMousemove);
+      d3.select(document).on("mouseup", onMouseup);
+      d3.select(document).on("keyup", function() {
+        if (d3.event.keyCode === ESCAPE_KEYCODE) {
+          reset(this);
+        }
+      });
+      el.on("touchmove", onMousemove);
+      d3.select(document).on("touchend", onMouseup);
+      d3.select(document).on("touchcancel", onMouseup);
+    }
+  }
+})
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -92,6 +92,10 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
     $scope.refreshGraph();
   });
 
+  $scope.$on('setRangeNoRefresh', function(ev, range) {
+    $scope.graph.range = range;
+  });
+
   $scope.$on('setRange', function(ev, range) {
     $scope.graph.range = range;
     $scope.refreshGraph(500);
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -123,15 +123,15 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       if (r === 1) {
         r = 0.5;
       }
+      var rangeSeconds = Prometheus.Graph.parseDuration($scope.graph.range);
+      var endTime = Math.floor($scope.graph.endTime / 1000);
       var scalingFactor = $(".widget_wrapper").outerWidth() * r / 10;
-      var rangeSeconds = Prometheus.Graph.parseDuration(scope.graph.range);
       // bigger denominator == smaller step == more data
       var step = Math.floor(rangeSeconds / scalingFactor)
-
       // Cancels the reload request if it exists.
       $timeout.cancel(debounce);
       debounce = $timeout(function() {
-        refreshFn(rangeSeconds, step).then(function(data) {
+        refreshFn(endTime, rangeSeconds, step).then(function(data) {
           scope.$broadcast('redrawGraphs', data);
           AnnotationRefresher(scope.graph, scope);
         });
--- a/app/assets/javascripts/angular/directives/graph_chart.js
+++ b/app/assets/javascripts/angular/directives/graph_chart.js
@@ -1,4 +1,4 @@
-angular.module("Prometheus.directives").directive('graphChart', ["$location", "WidgetHeightCalculator", "VariableInterpolator", "RickshawDataTransformer", "YAxisUtilities", function($location, WidgetHeightCalculator, VariableInterpolator, RickshawDataTransformer, YAxisUtilities) {
+angular.module("Prometheus.directives").directive('graphChart', ["$location", "$rootScope", "WidgetHeightCalculator", "VariableInterpolator", "RickshawDataTransformer", "YAxisUtilities", function($location, $rootScope, WidgetHeightCalculator, VariableInterpolator, RickshawDataTransformer, YAxisUtilities) {
   return {
     scope: {
       graphSettings: '=',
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -11,6 +12,145 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
       var graphData = [];
       var annotationData = [];
 
+      // figure out how to move this into a separate service, and
```
grobie: Mark a todo with `// TODO(stuart)`.

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -123,7 +123,8 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       var rangeSeconds = Prometheus.Graph.parseDuration(scope.graph.range);
       // bigger denominator == smaller step == more data
       var step = Math.floor(rangeSeconds / scalingFactor)
-      refreshFn(rangeSeconds, step).then(function(data) {
+      var endTime = Math.floor($scope.graph.endTime / 1000);
```
grobie: What is the reason the endTime calculation needs to be moved here? Not that I think it's a bad idea to not always use the global $scope in all methods (adds so many side effects), more for my code understanding of the change of this PR.

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -134,24 +276,24 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
...
-                scaleID = s.axisID;
-                break;
+                graphMax = axesBounds[matchingAxis.id].max;
+                bound.max = axesBounds[matchingAxis.id].max;
+              } else {
+                bound.max = maybeYMax;
+                graphMax = maybeYMax;
+              }
+              scaleID = s.axisID;
+              break;
```
grobie: All the line changes above look like broken indentation? Do you do these changes manual or is this your editor?

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -359,9 +503,9 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         if (scope.graphSettings.showLegend === "never" ||
             (scope.graphSettings.showLegend === "sometimes" && series.length > 5)) {
           $(element[0]).find(".legend").hide();
-          series.forEach(function(s) {
-            s.noLegend = true;
-          });
+        series.forEach(function(s) {
+          s.noLegend = true;
+        });
```
grobie: Again, broken indentation now?

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -11,6 +12,147 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
...
+          rectangle.remove()
+        }
+
+        function onMousemove() {
+          //Draw the rectangle
+          var limits = [drag.startPX, d3.event.offsetX].sort();
+          rectangle
+          .attr("fill", "steelblue")
+          .attr("x", limits[0])
+          .attr("width", limits[1]-limits[0]);
```
grobie: Looks like a very unusual way to indent a multi-line statement.

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -11,6 +12,147 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
...
+
+          drag.target = d3.event.target;
+          drag.startDt = pointAsDate(d3.event);
+          drag.y = d3.event.offsetY;
+          drag.startPX = d3.event.offsetX;
+          d3.event.preventDefault ? d3.event.preventDefault() : d3.event.returnValue = false;
+          d3.select(document).on("mousemove", onMousemove);
+          d3.select(document).on("mouseup", onMouseup);
+          d3.select(document).on("keyup", function() {
+            if (d3.event.keyCode === 27) {
```
grobie: Replace magic number with constant. http://www.refactoring.com/catalog/replaceMagicNumberWithSymbolicConstant.html

grobie:
This is awesome!!!

I believe it's important to reload data after zooming in. If I zoom in into a specific time range, it's because I want to inspect the data in more detail. Maybe it's possible to do: zoom in -> show existing data immediately -> trigger data reload -> replace graph with more detailed data from server.

Another open question is whether zooming in in one graph should also change the time range in all other graphs. I'd argue it should, because if I'm interested in a specific graph, e.g. CPU usage, it's very likely I also want to zoom in the same time range in another dashboard graph, e.g. memory usage. This means zooming in would change the dashboard time settings. Open for discussion of course.

app/assets/javascripts/angular/controllers/graph_ctrl.js
stuartnelson3: Originally, I was going to call `refreshFn()` (on the next line) with updated arguments coming from the user's drag selection. Currently I'm thinking I will just update the `$scope.graph.endTime` and `$scope.graph.range` variables and call for a refresh, though, which makes this unnecessary.

In general I'm attempting to pass variables explicitly instead of defaulting to sending in `$scope` and then picking off what attributes I need, for clarity and to ease the job of extracting the graph drawing into its own independent library (something that is eventually going to happen, so we can consolidate graph functionality between prometheus and promdash).

app/assets/javascripts/angular/directives/graph_chart.js
stuartnelson3: ah yeah, that's how my vim does it, and i was in "free-hack" mode, so i was just going with it

app/assets/javascripts/angular/directives/graph_chart.js
stuartnelson3: ah sheet, yeah this is completely unrelated, my editor must have picked this up and messed it up.

stuartnelson3:
Hmmmm, that seems like it could be a big assumption, especially given that there's (currently) no fast way to back out. Additionally, triggering a reload on all graphs could be slow. You're the main voice in the discussion, though, so it is mostly what solves your use-case.

stuartnelson3:
getting some weird behavior when dragging across some "magic" boundary where the box stops rendering, will investigate more tomorrow.

app/assets/javascripts/angular/services/drag_selection.js
```diff
@@ -0,0 +1,110 @@
...
+    d3.select(document).on("mousemove", onMousemove);
+    d3.select(document).on("mouseup", onMouseup);
+    d3.select(document).on("keyup", function() {
+      if (d3.event.keyCode === ESCAPE_KEYCODE) {
+        reset();
+      }
+    });
+    d3.select(document).on("touchmove", onMousemove);
+    d3.select(document).on("touchend, touchcancel", onMouseup); // i think i can listen to both events like this..
+    // d3.select(document).on("touchcancel", onMouseup);
```
grobie: Can you? If yes, cleanup comment?

stuartnelson3: need to get my hands on an ipad at work, would love to test using touch for zooming :)

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -135,6 +138,9 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
   }
 
   $scope.refreshGraph();
+
+  $scope.drag = new DragSelection();
```
stuartnelson3: this doesn't need to be created here; move it into the graph_chart and instantiate it there. also move the `setup()` function to the instantiation.

app/assets/javascripts/controllers.js
```diff
@@ -17,6 +17,27 @@ Prometheus.Graph = {
     's': 1
   },
 
+  durationToString: function(seconds) {
+    // TODO(stuart): this is always returning seconds
+    // look to round to a certain amount based on the units
+    // in the graph? some level of precision lost.
```
stuartnelson3: This is up to debate. Currently, I have yet to see a range defined by dragging a selection on the graph return anything other than seconds. I'm all for removing this function and always returning seconds above (app/assets/javascripts/angular/resources/shared_graph_behavior.js:86). Up to you guys.

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -277,6 +278,19 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         rsGraph.series.legend = legend;
         rsGraph.render();
 
+        var drag = new Rickshaw.Graph.DragZoom({
+          graph: rsGraph,
+          opacity: 0.5,
+          fill: 'steelblue',
+          minimumSelection: 25,
```
stuartnelson3: config is up to our preference. minimum selection is the number of pixels that a selection has to exceed in order for the highlighted area to be rendered. this protects against double clicking an area and having it render essentially nothing.

app/assets/javascripts/Rickshaw.Graph.DragZoom.js
```diff
@@ -0,0 +1,131 @@
+Rickshaw.namespace('Rickshaw.Graph.DragZoom');
```
stuartnelson3: once this is all code reviewed and we're happy with how it works i'm planning on submitting it to the rickshaw repo.

stuartnelson3:
@juliusv @grobie find some bugs for me

app/assets/javascripts/controllers.js
grobie: Can you explain this a bit more? If I understand you correctly, it's possible that the input is given in years, days, etc. (month is not possible?)? 

Where does this input come from? Is this a rickshaw thing or related to promdash?

stuartnelson3: the input is given in seconds as a large integer. Then, that number is `%`'d against the seconds value to see if it is perfectly divisible by it. If so, then the value returned will be in those units. otherwise, it will be in seconds.
example:

`durationToString(259200)` => '3d'
259200 % 60*60*24 (this is `factors['d']`) === 0, so it is evenly divisible by days

it is used at app/assets/javascripts/angular/resources/shared_graph_behavior.js:86, seen in this PR.

The input comes from the rickshaw graph. The seconds value is the range in seconds highlighted by the user. Rickshaw manages X axis times behind the scenes in seconds.

grobie:
:heart_eyes: This is so awesome. Super smooth experience, high five!

Now we need to talk about zooming out :D

One option would be to implement #289. This will essentially allow to maintain a history in the browser, so that `backspace` can be used to revert a step; therefore zooming out on backspace.

Another option would be to implement #147 adding for example a `z` shortcut to zoom out / revert an action.

stuartnelson3:
Do you want to do that in this PR, or in a separate one?

I have an idea, but it exists externally to the code in this PR.

app/assets/javascripts/angular/directives/graph_chart.js
grobie: Woud it be possible to use seconds as the minimum selection unit?

It's currently possible to zoom in into a total range <1s with values in ms as unit in the x axis. I guess it doesn't make sense to have a total range smaller than say 5s. Using time as absolute minimum instead of pixel width which is relative to the current state ensures zooming determines.

stuartnelson3: makes sense to me. I'll add the functionality and see what @juliusv thinks should be our minimum allowed zoom window

stuartnelson3:
ping, opinion?

juliusv:
Is there harm in allowing zoom windows all the way down to 1s?

stuartnelson3:
my only concern is there wouldn't be valuable data at such a small time range

juliusv:
Depends on the use case. Some people might run with 1s scraping. If it doesn't hurt, I wouldn't forbid it. They can always zoom out :)

stuartnelson3:
zooming out via the "back" button may be fixed implicitly by #292

grobie:
I'm vary about allowing such a small zoom range. It doesn't seem to make any sense with a default prometheus configuration (by default scrape interval is in the rang of 15s to 60s. Such a prometheus won't return any data points for a range of <5s). Though, I don't feel strong that we therefore have to forbid it.

:+1: if we're fine with back button / browser navigation for zooming out again. That can be discussed in the separate PR.

app/assets/javascripts/controllers.js
```diff
@@ -17,6 +17,27 @@ Prometheus.Graph = {
     's': 1
   },
 
+  durationToString: function(seconds) {
+    // TODO(stuart): this is always returning seconds
+    // look to round to a certain amount based on the units
```
juliusv: Hmm, I can't parse this sentence somehow, or is a new one? (please use sentence-style caps and punctuation)

juliusv:
This is magic and rocket science, I love it. Make it so! :+1: 

As for the minimum zoom, I don't feel strongly, so maybe 15s as a compromise is good.

app/assets/javascripts/controllers.js
stuartnelson3: Ah sorry, I'll fix this. Basically, in broken no good engrish, it's saying "This function, for where it is being used, will always return a duration in seconds. Look at rounding to the nearest value." I don't think this is necessarily valid that we want to round, and it should be pretty self-evident that only perfectly divisible time units will be days/weeks/years, so I'm going to delete the comment.

stuartnelson3:
I deferred to you regarding changing to 1 second, but my original reasoning follows what @grobie said: because of the standard scrape time, I really don't see the point in zooming in less than 15sec. I'll set it there, and make a note of it in the release to see if the customers have a strong opinion.

stuartnelson3:
Also, regarding zoom-out: I think ctrl+z would be a good option, and I think that should be implemented whether or not selecting "back" in the browser works. But, as you said, discussion for another pr.

stuartnelson3:
@juliusv do you have a preference on how to squash this? I'm thinking 2 commits: the first commit where I refactor the graphRefresher, then the rest of the commits (which is actually delightfully simple, as 99% of the code exists in a separate file).

app/assets/javascripts/Rickshaw.Graph.DragZoom.js
stuartnelson3: Note: I need to get some time with an iPad to get the touch swipe functionality working.

@juliusv are there any company ipads on the VPN (so they can access promdash)? it feels SO natural to look at a dashboard graph on an ipad, and swipe to zoom in. I would love to see if there's interest in adoption.

juliusv: @stuartnelson3 I don't think so, but this would be an internal matter, best discussed maybe with mobile teams :) In general, I'd be surprised if there were many orgs out there that allowed sufficiently privileged VPN on iPads to use monitoring, but who knows...

juliusv:
Yeah, I think that squashing strategy is fine.

juliusv:
(and yes, zoom -> different PR)

app/assets/javascripts/Rickshaw.Graph.DragZoom.js
grobie: @stuartnelson3 What is the state on contributing this back to rickshaw? Still planned?

stuartnelson3: still planned, but i haven't gotten around to it. will try to do soon.

Issue #153, opened 2014-04-28
Time range change feedback


matthiasr:
Currently, there is no visual feedback after clicking the +/- buttons to change the time range, until all the graphs are updated. I tend to click several times then and wildly overshoot.

As a minimal feedback, the time displayed should change before any potentially long-running action is taken. Even better would be to then start by scaling all the graphs, even if there is no new data to fill them in– then, the waiting time until this data is read and rendered will even contribute to a "smooth" transition; seeing the current graphs "shrink" should give a visual cue to the zooming "motion".

grobie:
Solved with #290.

Issue #287, opened 2015-01-22
debounce refresh on widgets after global config changes


stuartnelson3:
global config changes should have a debounce of ~200ms (or so) to prevent wasted requests to the prometheus backend if the user is rapidly updating date/range global settings.

stuartnelson3:
closed by #290

PR #290, opened 2015-01-23
set 500ms debounce timer on graph refresh

stuartnelson3:
requests to refresh are now queued in a timer. if an additional request is received before the timer expires, the original timer is canceled and a new one is created.

canceling an expired timer has no negative effects.

the 500ms debounce is arbitrary. Let me know what you guys think it should be set to. I originally tried 200ms but it was too quick, and would often reload between clicks.

@juliusv @grobie 

requested in #287

### DIFF
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -1,6 +1,7 @@
 angular.module("Prometheus.controllers").controller('GraphCtrl',
                                                     ["$scope",
                                                       "$http",
+                                                      "$timeout",
                                                       "GraphRefresher",
                                                       "YAxisUtilities",
                                                       "SharedWidgetSetup",
@@ -8,6 +9,7 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
                                                       "AnnotationRefresher",
                                                       function($scope,
                                                                $http,
+                                                               $timeout,
                                                                GraphRefresher,
                                                                YAxisUtilities,
                                                                SharedWidgetSetup,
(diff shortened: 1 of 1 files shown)


grobie:
This should fix #153. @matthiasr

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -123,10 +126,15 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       var rangeSeconds = Prometheus.Graph.parseDuration(scope.graph.range);
       // bigger denominator == smaller step == more data
       var step = Math.floor(rangeSeconds / scalingFactor)
-      refreshFn(rangeSeconds, step).then(function(data) {
+
+      // Cancels the reload request if it exists.
+      $timeout.cancel(debounce);
+      debounce = $timeout(function() {
+        refreshFn(rangeSeconds, step).then(function(data) {
         scope.$broadcast('redrawGraphs', data);
```
juliusv: missing indent for these lines?

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -123,10 +126,15 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
...
         AnnotationRefresher(scope.graph, scope);
-      });
+        });
+      }, 500);
```
juliusv: What concerns me is that this 500ms debounce is now added to any and all graph loads, not just the ones that are triggered by time adjustments. This includes initial page load, as well as any other manual changes to graphs which don't profit from being debounced. One solution I can think of is to pass a boolean `debounce` argument to `refreshGraph()`, which would only be set to `true` when the refresh originates from a time settings change, namely in these lines:

https://github.com/prometheus/promdash/blob/debounce/app/assets/javascripts/angular/controllers/graph_ctrl.js#L94-L102

stuartnelson3:
roger, debounce updated to only delay on changes to global time/date settings

Issue #281, opened 2015-01-17
Persist legend selection in short URLs


grobie:
Use case:
I want to send a link to a graph with the same legend settings I'm looking at, e.g. to highlight a 5xx response rate which would be barely visible otherwise.

stuartnelson3:
Can you clarify whether this is a link to a single widget, a dashboard with all the settings serialized into the URL, or both?

I'm guessing that in order to do this we would have to copy the legend selection data, which we're currently storing somewhat ethereally in memory during a redraw, to the graph data data-structure. if you want this functionality on a single widget page, it will definitely have to be persisted in the dashboard's JSON blob, and if you want it to be encoded in a URL, I could probably append it as data to each graph widget.

@grobie is this enough of a feature to warrant the additional code complexity, or would it be saving you clicking 1-3 legend items a couple times a day?

@juliusv thoughts?

juliusv:
Yeah, should be as straightforward as storing the legend state in the graph's settings instead of in an ephemeral local variable. Ok from my side, unless I overlooked something and it gets too complicated in the details.

stuartnelson3:
the only issue I can foresee is if the user saves after making this change. we'll have to update the JSON schema to include this new information. i assume this is ok?

juliusv:
Ok, I assumed that was what grobie wanted anyways, but maybe he was only talking about the single-widget graph link view. Guess I'm fine with doing it as I suggested though...

grobie:
I was only thinking about the single widget graphs **. I don't mind if it's possible to persist them in dashboards to, but I don't have a use case for this right now.

** I'm confused by the terminology in promdash, judging by the table name, I thought you name these "Short URLs"?

stuartnelson3:
closed by #291

PR #291, opened 2015-01-23
store graph legend state

stuartnelson3:
requested in #281 

@grobie @juliusv

### DIFF
--- a/app/assets/javascripts/angular/directives/graph_chart.js
+++ b/app/assets/javascripts/angular/directives/graph_chart.js
@@ -11,6 +11,8 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
       var graphData = [];
       var annotationData = [];
 
+      element.on("click", ".legend .line", storeLegendState);
+
       function setLegendString(series) {
         // TODO(stuartnelson3): Do something with this function. Put it somewhere or simplify it.
         var expressions = scope.graphSettings.expressions;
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -11,6 +12,8 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
       var graphData = [];
       var annotationData = [];
 
+      element.on("click", ".legend .line", storeLegendState);
```
stuartnelson3: Previously, storing legend state was tied to manually refreshing the graph, as that was where graph legend state was lost. However, changing the legend state only triggers an update within rickshaw, not within our code, so the it was possible for a user to change the legend state, request a shortlink, and upon visiting the shortlink find that the legend state wasn't persisted.

now, any time the legend is changed, that state is stored.

juliusv:
Ah, interesting. Am I right in seeing that most of the complexity of this PR comes because we only want to store the legend state for shortlinks, and not always in the graph JSON? Otherwise we could just always store it in the normal graph settings and wouldn't need an additional parameter to the directives? Are there any good arguments for not always saving it? If some viewer deselects a time series and saves the dashboard, the next viewer will also see that time series as deselected, but that could be a feature, not a bug?

grobie:
Fine with me. If accidental saves cause confusion, we can always separate
edit+show views more ;)

stuartnelson3:
Yeah all the complexity exists because I was trying to avoid saving the legend state. I'm totally fine with doing this, though, if you think it would be a good feature. I'll throw that in on monday, it'll simplify this immensely.

stuartnelson3:
k, updated to store legend state for graphs

grobie:
:+1: (maybe there is a more expressive name than `disabledSeries`, but I don't have one either)

app/assets/javascripts/angular/directives/graph_chart.js
```diff
@@ -389,6 +384,15 @@ angular.module("Prometheus.directives").directive('graphChart', ["$location", "W
         return "<table class=\"labels_table\">" + labelRows.join("") + "</table>";
       }
 
+      function storeLegendState() {
+        if (!rsGraph) {
+          return;
+        }
+        rsGraph.series.forEach(function(s) {
```
juliusv: Is this repeating that pattern from the previous PR where we store all the state instead of just the disabled ones in the map? Please don't do that, as there can be gazillions of series in a single dashboard, and we don't want to store them all in the graph JSON, only the disabled ones, if any.

stuartnelson3: ah, right. i made this change because the previous implementation never removed the series, but now i see i just need to zero it out each refresh. good catch.

Issue #204, opened 2014-09-30
React to URL changes


grobie:
> As a user,
> when I change a until/range/var.xxx URL parameter in the address bar,
> then I expect the graphs to update accordingly immediately.

grobie:
Solved in the meantime.

PR #292, opened 2015-01-23
Changes made to url vars and range/until propagate to the widgets

stuartnelson3:
No longer need to hit enter twice

the reverse process will come in a separate PR, changes to url vars on a dashboard => url in location bar

requested in #204 

@grobie @juliusv


grobie:
Existing template variables don't load anymore with this change. I also can't add new ones. Example is the our internal `bazooka-service-template` dashboard.

stuartnelson3:
Existing template variables are working for me in bazooka-service-template. Can confirm that adding new template variables wasn't working, accidentally removed that function (has been added back).

juliusv:
:+1:

PR #294, opened 2015-01-26
Add support for new query result JSON format.

juliusv:
See https://github.com/prometheus/prometheus/issues/389.

This is a transient change and the "||" switches will be removed after a
week or so, when no old-format Prometheus servers are in use anymore.


juliusv:
@stuartnelson3

stuartnelson3:
:+1:

Issue #288, opened 2015-01-23
Error encoding an entiry dashboard in URL


grobie:
Encoding an entire dashboard in base64 can easily exceed 4096 bytes which is a typical limit for URLs. 

As this makes these links useless, I suggest to save the dashboard URLs similar to graph URLs in database.

stuartnelson3:
if a user keeps the url updated, that would require an async request every time something in the dashboard changes to persist this, and it would also require some form of indicator in the URL that indicates to the UI to load from a particular place.

are you proposing to maintain the current functionality, but back the data in the database? or would it be a button somewhere, like "get link for this dashboard", and then it generates the link (pretty much exactly like the graph link)

juliusv:
AFAIK the max length of the hash part (fragment identifier) is unspecified
because it's not actually sent over HTTP, but only used locally.

I tried storing 1MB in the hash in Firefox and it worked fine to retrieve
it. Chrome's UI gets a bit sluggish at those sizes, but at least I was able
to verify that I was able to store 32k in the hash in Chrome. So maybe this
isn't really a big problem so far.

stuartnelson3:
close-able?

juliusv:
I'd wait to see what @grobie thinks :)

If the hash does turn out to be a problem, we could simply offer a
"permalink" button that only generates a permalink when requested instead
of constantly creating it upon every change.

grobie:
I can't paste such URL in irc, it fills so many lines in emails that I
neither feel comfortable to use it there and it's also too long for
SoundCloud's internal link shortener. This makes this feature completly
useless to me. While I don't think it needs a high priority, I consider
this to be a problem.

A permalink button sounds fine, no need for dynamic updates. I started to
clone dashboards as a workaround.
I'd wait to see what @grobie thinks :)

If the hash does turn out to be a problem, we could simply offer a
"permalink" button that only generates a permalink when requested instead
of constantly creating it upon every change.

On Fri, Jan 23, 2015 at 1:06 PM, stuart nelson <notifications@github.com>
wrote:

> close-able?
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/prometheus/promdash/issues/288#issuecomment-71184691>.

>

--
Reply to this email directly or view it on GitHub
<https://github.com/prometheus/promdash/issues/288#issuecomment-71191594>.

juliusv:
Ok, agreed.

stuartnelson3:
closed by #296

PR #296, opened 2015-01-26
offer dashboard perma-link instead of base64 encoding after hash

stuartnelson3:
requested in #288 

question: keep url updated is still an option. do we want this to be removed?

### DIFF
--- a/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
@@ -28,6 +28,24 @@ angular.module("Prometheus.controllers")
     $scope.$broadcast('redrawGraphs');
   }
 
+  $scope.generatePermalink = function(event) {
+    if ($scope.generatingPermalink || $scope.showPermalink) {
+      $scope.showPermalink = false;
+      return;
+    }
+    $scope.generatingPermalink = true;
+    $http.post($window.location.origin + "/permalink", buildDashboardJSON()).then(function(payload) {
+      $scope.showPermalink = true;
+      $scope.permalink = $window.location.origin + payload.data.url;
+      var input = $("[ng-model=permalink]")[0];
+      InputHighlighter(input);
+    }, function() {
+      alert("Failed to create dashboard permalink");
+    }).finally(function() {
+      $scope.generatingPermalink = false;
+    });
+  };
+
   $scope.showCloneControls = false;
   $scope.fullscreen = false;
   $scope.saving = false;
@@ -51,14 +69,7 @@ angular.module("Prometheus.controllers")
 
   $scope.saveDashboard = function() {
     $scope.saving = true;
-    $http.put(window.location.pathname + '.json', {
-      'dashboard': {
-        'dashboard_json': {
-          'globalConfig': $scope.globalConfig,
-          'widgets': $scope.widgets
-        }
-      }
-    }).error(function(data, status) {
+    $http.put($window.location.pathname + '.json', buildDashboardJSON()).error(function(data, status) {
       alert("Error saving dashboard.");
     }).finally(function() {
       $scope.saving = false;
@@ -84,6 +95,7 @@ angular.module("Prometheus.controllers")
         $scope.closeCloneControls();
         $scope.$broadcast('closeTabs');
         $scope.showDashboardSettings = false;
+        $scope.showPermalink = false;
       });
     }
   });
@@ -185,17 +197,6 @@ angular.module("Prometheus.controllers")
     } else {
       $location.search("until", null);
     }
-    if (newConfig.encodeEntireURL) {
-      URLConfigEncoder({globalConfig: newConfig, widgets: $scope.widgets});
-    } else {
-      $location.hash(null);
-    }
-  }, true);
-
-  $scope.$watch("widgets", function(newWidgets, oldWidgets) {
-    if ($scope.globalConfig.encodeEntireURL) {
-      URLConfigEncoder({globalConfig: $scope.globalConfig, widgets: newWidgets});
-    }
   }, true);
 
   $scope.$watch('globalConfig.tags', function() {
@@ -244,4 +245,16 @@ angular.module("Prometheus.controllers")
   $scope.copyWidget = function() {
     $scope.widgets.push(angular.copy($scope.widgetToClone));
   };
+
+  function buildDashboardJSON() {
+    return {
+      'dashboard': {
+        'name': dashboardName,
+        'dashboard_json': {
+          'globalConfig': $scope.globalConfig,
+          'widgets': $scope.widgets
+        }
+      }
+    };
+  }
 }]);
--- a/app/assets/templates/frame_template.html
+++ b/app/assets/templates/frame_template.html
@@ -3,7 +3,7 @@
     <div class="widget_title panel-heading" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
       {{getTitle()}}
       <div class="btn-group pull-right graph_config_menu" ng-show="showControls || showTab">
-        <div class="btn btn-primary btn-sm" ng-click="refreshFrame()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestsInFlight}" class="icon-cycle"></i></div>
+        <div class="btn btn-primary btn-sm" ng-click="refreshFrame()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestsInFlight}" class="glyphicon glyphicon-refresh"></i></div>
         <div class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'frame_source')" tooltip-append-to-body="true" tooltip="Frame Source"><i class="icon-add-to-list"></i></div>
         <div class="btn btn-primary btn-sm" ng-show="frame.graphite" ng-click="toggleTab($event, 'timerange')" tooltip-append-to-body="true" tooltip="Time options"><i class="icon-clock"></i></div>
         <div class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'frame_settings')" tooltip-append-to-body="true" tooltip="Frame Settings"><i class="icon-cog"></i></div>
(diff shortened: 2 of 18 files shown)


spec/features/dashboards_spec.rb
```diff
@@ -29,6 +29,24 @@
     end
   end
 
+  describe "static dashboards" do
+    scenario "creating a static dashboard" do
+      dashboard = FactoryGirl.create(:dashboard)
+      visit dashboard_slug_path dashboard.slug
+      expect {
+        find('.glyphicon-link').click
```
stuartnelson3: thoughts on this? it's just an icon so i'm targeting a class, but using a generic class is a bad idea (even if right now it is unique), but i'm not sure how i feel about adding a testing-specific class.

app/views/dashboards/show.html.erb
```diff
@@ -65,7 +65,32 @@
...
+            <span ng-show="generatingPermalink" class="glyphicon glyphicon-refresh spin"></span>
+          </button>
+        <% end %>
       </div>
+      <% unless @dashboard.static %>
+        <div class="dashboard_settings panel panel-default" ng-show="showPermalink">
+          <div class="panel-heading">Dashboard Permalink</div>
+          <div class="panel-body">
+            <div class="input-group">
+              <input type="text" class="form-control" ng-model="permalink">
```
stuartnelson3: the width on this is weird. i'm calling in @juliusv to fix it since he owes me some css work :P

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
     $scope.$broadcast('redrawGraphs');
   }
 
+  $scope.generateStaticDashboard = function(event) {
+    // TODO: When the input with the static dashboard link is visible,
+    // should clicking the "generate" button close the input, or
+    // generate a NEW static link?
```
stuartnelson3: just calling attention to the above comment

app/views/dashboards/show.html.erb
```diff
@@ -65,7 +65,32 @@
         <button class="btn btn-default" ng-click="enableFullscreen()" title="Fullscreen mode">
           <span class="glyphicon glyphicon-fullscreen"></span>
         </button>
+        <% unless @dashboard.static %>
+          <button class="btn btn-default" ng-click="generateStaticDashboard()"
+            tooltip-append-to-body="true" tooltip="Create dashboard permalink">
```
juliusv: I'd go for even more explicit "Create permalink to current dashboard state"

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
juliusv: I think closing would be both the better and more consistent thing... haven't read the rest of the code yet, but one question would be whether we can detect if anything has changed and only regenerate a link in that case (as we do with individual graph permalinks, I think)

app/views/dashboards/show.html.erb
juliusv: Or maybe "Get permalink to current dashboard state"

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
...
+    $scope.generatingPermalink = true;
+    $http.post(window.location.origin + "/static", {
+      'dashboard': {
```
juliusv: Seems like this whole JSON object is now duplicated from where we do the normal dashboard saving. Can we refactor it into a `buildDashboardJSON()` function?

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
...
+        'name': dashboardName,
+        'dashboard_json': {
+          'globalConfig': $scope.globalConfig,
+          'widgets': $scope.widgets
+        }
+      },
+    }).then(function(payload) {
+      $scope.showPermalink = true;
+      $scope.permalink = window.location.origin + payload.data.url;
```
juliusv: Not `$window`?

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
...
+      var input = $("[ng-model=permalink]")[0];
+      InputHighlighter(input);
+    }, function() {
+      console.log("failed")
```
juliusv: "Failed to create dashboard permalink"

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
     $scope.$broadcast('redrawGraphs');
   }
 
+  $scope.generateStaticDashboard = function(event) {
+    // TODO: When the input with the static dashboard link is visible,
+    // should clicking the "generate" button close the input, or
+    // generate a NEW static link?
+    $scope.generatingPermalink = true;
```
juliusv: perhaps add a line before this that just returns when the link is already currently being generated?

app/assets/templates/frame_template.html
```diff
@@ -3,7 +3,7 @@
     <div class="widget_title panel-heading" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
       {{getTitle()}}
       <div class="btn-group pull-right graph_config_menu" ng-show="showControls || showTab">
-        <div class="btn btn-primary btn-sm" ng-click="refreshFrame()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestsInFlight}" class="icon-cycle"></i></div>
+        <div class="btn btn-primary btn-sm" ng-click="refreshFrame()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestsInFlight}" class="glyphicon glyphicon-refresh"></i></div>
```
juliusv: There are a couple more icons which look way better in Glyphicons - like the permalink icon and the trash bin. But let's keep that for a future PR :)

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
stuartnelson3: :+1:

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
stuartnelson3: ah hah there are several errant `window`s that can be changed

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
stuartnelson3: console.log or alert? i'm thinking alert, as that's how we inform the user a save of the dashboard failed

app/controllers/static_dashboards_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  # Use callbacks to share common setup or constraints between actions.
+  def set_dashboard
+    @dashboard = Dashboard.find(params[:id])
+  end
+
+  def set_directories
+    @directories = Directory.all
+  end
+
+  # Never trust parameters from the scary internet, only allow the white list through.
```
juliusv: Seems like duplicating some complex code. I'm wondering whether to have just one controller for both types of dashboard and just send the "static" attribute as part of the normal POST data instead of differentiating it in the URL.

app/views/dashboards/show.html.erb
juliusv: fine for now, doesn't look too bad :P

app/views/dashboards/show.html.erb
```diff
@@ -65,7 +65,32 @@
...
+                  tooltip-append-to-body="true"
+                  tooltip="Open dashboard permalink in new tab" target="_blank"
+                  ng-click="showPermalink = false">
+                  <span class="glyphicon glyphicon-new-window"></span>
+                </a>
+              </span>
+            </div>
+          </div>
+        </div>
+      <% end %>
```
juliusv: can't comment on the right line, but: shouldn't the "Save" button also disappear for permlink dashboards? At least pressing it doesn't work, and I'd recommend hiding it.

spec/features/dashboards_spec.rb
juliusv: Don't have a much better idea, but I'd at least search only in `#global_controls` for this glyphicon.

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -30,6 +30,31 @@ angular.module("Prometheus.controllers")
...
+    $http.post(window.location.origin + "/static", {
```
juliusv: I was confused by the terminology "static" vs. "dynamic" and assumed it meant the other way around: dashboards generated "dynamically" by clicking the permalink button vs. "static" ones which are the "real" ones :)

How about e.g. "/perm" for the URLs and "permalink" for the boolean DB column? Now just needs a good name for the "normal" dashboards :)

juliusv:
As for keeping the option to encode the entire dashboard in the URL, I vote for removing it.

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -28,6 +28,25 @@ angular.module("Prometheus.controllers")
     $scope.$broadcast('redrawGraphs');
   }
 
+  $scope.generateStaticDashboard = function(event) {
+    // finish checking to see if it is a duplicate dashboard
```
juliusv: Sentence-style plz

stuartnelson3:
@juliusv ready for review

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -23,6 +23,14 @@
       expect(response).to redirect_to(dashboard_slug_path(d.slug))
     end
 
+    it "creates a new static dashboard" do
```
juliusv: s/static dashboard/dashboard permalink/ (dito below)

juliusv:
:+1: besides the last couple of "static" mentions. Feel free to squash-n-merge after fixing.

PR #299, opened 2015-01-28
Zoom out

stuartnelson3:
part two of the zoom spectacular

### DIFF
--- a/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
@@ -86,8 +86,19 @@ angular.module("Prometheus.controllers")
     $scope.nextCycleRedraw();
   };
 
+  $window.history.replaceState({initial: true}, $("title").text(), $window.location.href);
   $document.keydown(function(ev) {
-    if (ev.keyCode === 27) { // Escape keycode
+    var Z_KEY = 90;
+    var ESC_KEY = 27;
+    var validElement = ["INPUT", "TEXTAREA"].indexOf(document.activeElement.tagName) === -1;
+    if (ev.keyCode === Z_KEY && validElement) {
+      if (($window.history.state || {}).initial) {
+        return;
+      }
+      $window.history.back();
+      return;
+    }
+    if (ev.keyCode === ESC_KEY) {
       if ($scope.fullscreen) {
         $scope.exitFullscreen();
       }
--- a/app/assets/javascripts/angular/directives/frame.js.erb
+++ b/app/assets/javascripts/angular/directives/frame.js.erb
@@ -1,4 +1,6 @@
-angular.module("Prometheus.directives").directive('inlineFrame', function() {
+angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "VariableInterpolator", "GraphiteTimeConverter", "WidgetHeightCalculator", function($sce, VariableInterpolator, GraphiteTimeConverter, WidgetHeightCalculator) {
+  var FRAME_CONTAINER_PADDING = 10;
+  var FRAME_BACKGROUND_COLOR = '%23191919';
   return {
     scope: {
       frame: "=frameSettings",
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -83,6 +83,10 @@ angular.module("Prometheus.controllers")
   };
 
   $document.keydown(function(ev) {
+    if (ev.keyCode === 90 && ev.metaKey) { // Z keycode + cmd key
```
juliusv: As @grobie suggested elsewhere once, please extract magic numbers into nicely-named constants.

juliusv: Also: how do I zoom out on Linux? Ctrl-Z doesn't seem to work.

stuartnelson3: hm, i was thinking that would be an issue. will add support for ctrl+z as well.

grobie: I would recommend not to use any command/ctrl key combinations in js. It seems to cause more trouble than good in my experience, as people will use a huge set of different shortcut combinations.

Feel free to ignore me, but I'd vote to keep promdash shortcuts simple and similar to gmail's shortcuts, so just `z` here.

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -8,6 +8,64 @@ angular.module("Prometheus.directives").directive('inlineFrame', function() {
...
+      });
+
+      // In order for zoom-out to work correctly on dashboards with iframes,
+      // the iframe needs to be completely reconstructed on each refresh.
+      // Modifying an iframe's src attribute, as in the previous implementation,
+      // changes the browser's history. This interrupts a user's attempt to zoom
+      // out as it takes an additional history.back() execution to traverse the
+      // history state added by modifying an iframe's src.
+      // http://khaidoan.wikidot.com/iframe-and-browser-history
+      function regenURL() {
```
juliusv: More like `regenFrame`, right?

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -8,6 +8,64 @@ angular.module("Prometheus.directives").directive('inlineFrame', function() {
...
+          } else {
+            targets.push(f);
+          }
+        });
+
+        fields['height'] = scope.frameHeight().height;
+        fields['width'] = scope.frameHeight().height / scope.aspectRatio;
+        fields['from'] = GraphiteTimeConverter.graphiteFrom(scope.frame.range, scope.frame.endTime);
+        fields['until'] = GraphiteTimeConverter.graphiteUntil(scope.frame.endTime);
+        fields['bgcolor'] = '%23191919'
```
juliusv: missing semicolon

app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -48,15 +50,14 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
         $scope.fullscreenTitle = true;
       }
 
-      if (urlVars.range) {
-        $scope.globalConfig.range = urlVars.range;
-        $scope.setRange()
-      }
+      $scope.globalConfig.range = urlVars.range || $scope.originalRange;
+      $scope.setRange()
```
juliusv: missing semi

app/assets/javascripts/angular/directives/frame.js.erb
stuartnelson3: good point

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
juliusv: Ah, the Gmail approach sounds good too. Of course the "z" should not do anything when you're typing it anywhere in a text field :)

stuartnelson3:
let me know what you think

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -82,8 +82,18 @@ angular.module("Prometheus.controllers")
     $scope.nextCycleRedraw();
   };
 
+  $window.history.replaceState({initial: true});
```
stuartnelson3: This records the initial spot in history that the user loads the page. Without using this check, the user could hit "z" to go back in their history to the page BEFORE the dashboard, which is undesired.

stuartnelson3:
@juliusv ready for review

stuartnelson3:
@juliusv after a lengthy rebase, ready for review. but feel free to leave it for monday..

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -74,8 +74,18 @@ angular.module("Prometheus.controllers")
     $scope.nextCycleRedraw();
   };
 
+  $window.history.replaceState({initial: true}, "Promdash - " + dashboardName, $window.location.href);
   $document.keydown(function(ev) {
-    if (ev.keyCode === 27) { // Escape keycode
+    var Z_KEY = 90;
+    var ESC_KEY = 27;
+    if (ev.keyCode === Z_KEY && document.activeElement.tagName !== "INPUT") {
```
grobie: Is it necessary to check for input? Will a keypress really trigger the event otherwise? I'd have expected that the browser doesn't forward keypress events anyway in that situation.

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
```diff
@@ -74,8 +74,18 @@ angular.module("Prometheus.controllers")
     $scope.nextCycleRedraw();
   };
 
+  $window.history.replaceState({initial: true}, "Promdash - " + dashboardName, $window.location.href);
```
grobie: I might be wrong, but isn't that a different title we usually use? It seems the title set by Rails is "<dashboard> - Prometheus Dashboard". Can we read the current value here, that would avoid us from maintaining the title format in two different places.

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
stuartnelson3: yeah, it definitely bubbles up if an `<input>` element has focus

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
stuartnelson3: updated

stuartnelson3:
@juliusv ping

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -7,6 +7,73 @@ angular.module("Prometheus.directives").directive('inlineFrame', function() {
...
+        }
+        parser.search = parser.search + '&decache=' + scope.refreshCounter;
+        return parser.href;
+      }
+
+      function parseGraphiteURL(parser) {
+        var queryStringComponents = parser.search.substring(1).split('&');
+        var fields = {};
+        var targets = [];
+        var FRAME_CONTAINER_PADDING = 10;
```
juliusv: Since this is constant-ish, how about putting it at the top of the file? This way it's even clearer that it's meant to be a constant and no need to rebuild this variable.

juliusv: Then again, it only needs visibility in the frame-related code. Maybe instead, put it after the first line in the file, before returning the scope and stuff?

app/assets/javascripts/angular/controllers/dashboard_ctrl.js
juliusv: We should probably exempt textareas too. If we add them later (like if someone wants to have a larger expression input), we probably wouldn't notice during testing that pressing Z in a textarea zooms out, and only users would finally let us know.

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -7,6 +7,73 @@ angular.module("Prometheus.directives").directive('inlineFrame', function() {
...
+          } else {
+            targets.push(f);
+          }
+        });
+
+        fields['height'] = scope.frameHeight().height;
+        fields['width'] = (scope.frameHeight().height / scope.aspectRatio) - FRAME_CONTAINER_PADDING;
+        fields['from'] = GraphiteTimeConverter.graphiteFrom(scope.frame.range, scope.frame.endTime);
+        fields['until'] = GraphiteTimeConverter.graphiteUntil(scope.frame.endTime);
+        fields['bgcolor'] = '%23191919';
```
juliusv: Extract this into a "constant" as well?

juliusv:
:+1: besides comments

PR #300, opened 2015-01-28
Add layerX for firefox calculations

stuartnelson3:
drag-to-zoom wasn't working in firefox
@juliusv 

from http://stackoverflow.com/questions/22716333/firefox-javascript-mousemove-not-the-same-as-jquery-mousemove#answer-22718173:

>And they only have offsetX and offsetY properties on browsers that supports them. There was a ticket to make them available for Firefox too, but was closed as wontfix because of performance reasons.

nb: sorry tobi, i moved helper functions around cause they were getting in my way, and then lumped all the changes into one commit.

### DIFF
--- a/app/assets/javascripts/Rickshaw.Graph.DragZoom.js
+++ b/app/assets/javascripts/Rickshaw.Graph.DragZoom.js
@@ -39,10 +39,6 @@ Rickshaw.Graph.DragZoom = Rickshaw.Class.create({
     this.svg.on("mousedown", onMousedown);
     this.svg.on("touchstart", onMousedown);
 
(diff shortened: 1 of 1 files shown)


juliusv:
:+1: working for me in both Chrome and FF.

Regarding commit messages, I'd focus on the intent rather than the implementation details: E.g. "Fix zoom in Firefox." is much clearer to anyone reading the git log later what this commit is actually about.

stuartnelson3:
right! you keep telling me and i keep forgetting, sorry

juliusv:
You had the option to amend the commit message before merging ;)

PR #313, opened 2015-01-30
Update js tests

stuartnelson3:


grobie:
Yay for enabling all tests again!

The rest of the PR seems to be basically code cleanup/updates without any behavior or tests changes, right? :+1: in that case.

stuartnelson3:
correct, this a.) updates our version of jasmine, b.) correctly sets up our tests with `angular-mock.js`, and c.) uncomments an entire test file that had been commented out because i hadn't properly set up the tests.

juliusv:
Woohoo :+1:

PR #314, opened 2015-02-02
Add travis settings

stuartnelson3:
moving back to travis because semaphore isn't currently supporting phantomjs correctly


juliusv:
:+1:

PR #315, opened 2015-02-02
add missing dependency

stuartnelson3:
adds missing dependency that causes breakout graphs to break

### DIFF
--- a/spec/javascripts/angular/controllers/single_widget_ctrl_spec.js
+++ b/spec/javascripts/angular/controllers/single_widget_ctrl_spec.js
@@ -1,0 +1,18 @@
+var servers = [];
+var graphBlob = {};
+describe('SingleWidgetCtrl', function() {
+  var $controller;
+  var scope;
+  beforeEach(inject(function(_$controller_, $rootScope){
+    $controller = _$controller_;
+    scope =  $rootScope;
+  }));
+
+  it('loads correctly', function() {
+    var $scope = scope.$new(true);
+    var controller = $controller('SingleWidgetCtrl', { $scope: $scope });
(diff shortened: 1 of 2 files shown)


stuartnelson3:
@juliusv threat level 1..

juliusv:
:+1:

stuartnelson3:
currently failing because semaphore has an incorrect phantomjs version

spec/javascripts/angular/controllers/single_widget_ctrl_spec.js
```diff
@@ -0,0 +1,18 @@
...
+  beforeEach(inject(function(_$controller_, $rootScope){
+    $controller = _$controller_;
+    scope =  $rootScope;
+  }));
+
+  it('loads correctly', function() {
+    var $scope = scope.$new(true);
+    var controller = $controller('SingleWidgetCtrl', { $scope: $scope });
+    expect($scope.servers).toEqual([]);
+    expect($scope.globalConfig.aspectRatio).not.toBe(undefined);
```
grobie: Did these tests catch the issue? Bonus points for red-green testing (http://en.wikipedia.org/wiki/Test-driven_development#Development_style).

stuartnelson3: yeah they did, gave the same missing dependency error that was reported. this is really just a smoke test to make sure everything runs.

grobie: This is great! Ideally all bug reports will be first written as tests which fail and then fixed (red-green testing). :lollipop: work :)

PR #318, opened 2015-02-02
Add js specs

stuartnelson3:
low hanging fruit controller specs

### DIFF
--- a/spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
+++ b/spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
@@ -1,0 +1,32 @@
+describe('DashboardCtrl', function() {
+  var $controller;
+  var $scope;
+  beforeEach(inject(function(_$controller_, $rootScope){
+    $controller = _$controller_;
+    $scope = $rootScope.$new(true);
+  }));
+
+  describe('visual toggles', function() {
+    describe('fullscreen', function() {
+      beforeEach(function() {
+        $controller('DashboardCtrl', { $scope: $scope });
+      });
+
+      it('starts not in fullscreen', function() {
+        expect($scope.fullscreen).toEqual(false);
+      });
+
+      it('toggles to fullscreen', function() {
+        $scope.enableFullscreen();
+        expect($scope.fullscreen).toEqual(true);
+      });
+
+      it('leaves fullscreen', function() {
+        $scope.enableFullscreen();
+        expect($scope.fullscreen).toEqual(true);
+        $scope.exitFullscreen();
+        expect($scope.fullscreen).toEqual(false);
+      });
+    });
+  });
+});
--- a/spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
+++ b/spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
@@ -7,6 +7,29 @@ describe('DashboardCtrl', function() {
   }));
 
   describe('visual toggles', function() {
+    describe('clone modal', function() {
+      beforeEach(function() {
+        $controller('DashboardCtrl', { $scope: $scope });
+      });
+
+      it('starts closed', function() {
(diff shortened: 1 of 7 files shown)


spec/javascripts/angular/services/variable_interpolator_spec.js
```diff
@@ -41,4 +42,11 @@ describe('VariableInterpolator', function() {
     var formatStr = "{{quantile | toPercent}} for {{server | hostname}}";
     expect(variableInterpolator(formatStr, this.availableFields)).toEqual("75% for localhost:8080");
   });
+
+  describe("regex", function() {
+    it("replace", function() {
```
stuartnelson3: @juliusv covers the case you just told me about via IRC

spec/javascripts/angular/services/variable_interpolator_spec.js
```diff
@@ -8,7 +8,8 @@ describe('VariableInterpolator', function() {
     this.availableFields = {
       name: "field name",
       quantile: "0.75",
-      server: "http://localhost:8080/metrics"
+      server: "http://localhost:8080/metrics",
+      job: "prometheus-stream"
```
grobie: Is it valid to keep a trailing comma in the last line in JS? Then the diffs become cleaner.

spec/javascripts/angular/services/variable_interpolator_spec.js
grobie: :lollipop: for catching issues with tests :)

spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
```diff
@@ -0,0 +1,55 @@
...
+        $controller('DashboardCtrl', { $scope: $scope });
+      });
+
+      it('starts not in fullscreen', function() {
+        expect($scope.fullscreen).toEqual(false);
+      });
+
+      it('toggles to fullscreen', function() {
+        $scope.enableFullscreen();
+        expect($scope.fullscreen).toEqual(true);
```
grobie: Do you think it would make sense to add another validation here to verify that fulltext also changes the UI? At the moment the text basically covers that `enableFullscreen` and `exitFullscreen` change the `fullscreen` variable between true and false. It might be good to have a test somewhere to ensure that it removes the navigation bar, form elements and makes the graphs bigger or so.

grobie:
Great work!

spec/javascripts/angular/services/variable_interpolator_spec.js
stuartnelson3: it is, actually. updated.

spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
stuartnelson3: right. these are just unit tests, so they don't hook into the visuals.

we have two options for feature specs:

we can stick with capybara, or we can move to protractor (a node.js based solution). protractor gives finer control over checking the DOM (you can see in some of the tests for stz), but it's an additional dependency and I'm not sure how to interface between the protractor being run by node and ruby-side things (setting up servers and dashboards).

grobie: Is it necessary to get fine dom control? If there are no fullscreen feature specs, I'd vote to open an issue for that.

grobie:
:+1:

spec/javascripts/angular/controllers/dashboard_ctrl_spec.js
stuartnelson3: i personally don't think it's necessary, and staying with capybara should be fine.

there are no fullscreen feature specs. feel free to open an issue :)

juliusv:
:+1: TODO(tomorrow): meet with @stuartnelson3 to talk about how we can make the regex/pipe/var-interpolation stuff better.

PR #320, opened 2015-02-03
Fix full screen aspect ratio

stuartnelson3:
depends on #299


app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -1,6 +1,6 @@
 angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "$timeout", "$location", "URLConfigDecoder", "URLVariablesDecoder", "ThemeManager", function($http, $timeout, $location, URLConfigDecoder, URLVariablesDecoder, ThemeManager) {
-  function commonSetup($scope) {
-    $scope.globalConfig = dashboardData.globalConfig || {
```
stuartnelson3: this was squashing any settings made to `$scope.globalConfig` before executing `commonSetup`. so, now the config object is explicitly passed in. this file and many others would benefit from refactoring.

juliusv:
:+1:

PR #321, opened 2015-02-03
Shared graph spec

stuartnelson3:
fixes bug where globalConfig endtime/range squash individually set widget endtime/range on initial load.

@juliusv

PR #322, opened 2015-02-03
Drag zoom in breakout

stuartnelson3:
drag-zoom in a breakout graph currently correctly zooms in on the dragged area, but the endTime and range are not updated. when the request is fired for more detailed visuals, it is sent with the old endTime and range, which causes the graph to return to its pre-zoomed state.

this fixes that behavior, as well as testing that query params for range and until are correctly removed and added to the url when the global config object is changed. changing the global config object is part of the callback from the drag zoom, and the changes propagate from there to the url.

depends on #321


juliusv:
:+1:

PR #323, opened 2015-02-04
Jshint

stuartnelson3:
jshint the code, and have jshint run on travis

@juliusv

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ before_script:
   - mysql -e 'create database promdash_test;'
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+  - npm install jshint
 script:
   - cp config/database.yml.example config/database.yml
   - bundle exec rake db:migrate
   - bundle exec rake spec
   - bundle exec rake spec:javascript
+  - jshint app/assets/javascripts
--- a/README.md
+++ b/README.md
@@ -6,11 +6,26 @@ Dashboard builder for Prometheus.
 
 [![Build Status](https://travis-ci.org/prometheus/promdash.svg?branch=master)](https://travis-ci.org/prometheus/promdash)
 
-    # Ruby tests.
-    bundle exec rake
+### Ruby
 
-    # Javascript tests.
-    RAILS_ENV=test bundle exec rake spec:javascript
+```bash
+$ bundle exec rake
+```
+
+### Javascript
+
+```bash
+$ RAILS_ENV=test bundle exec rake spec:javascript
+```
+
+### JSHint
+JSHint is run against all pull requests. To shorten the feedback loop, it is recommended that you run JSHint locally.
+With [NPM installed](https://docs.npmjs.com/getting-started/installing-node), install and run JSHint:
+
+```bash
+$ npm install jshint -g
+$ jshint app/assets/javascripts
+```
 
 ## Deployment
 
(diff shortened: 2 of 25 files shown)


README.md
```diff
@@ -6,12 +6,23 @@ Dashboard builder for Prometheus.
...
 [![Build Status](https://travis-ci.org/prometheus/promdash.svg?branch=master)](https://travis-ci.org/prometheus/promdash)
 
-    # Ruby tests.
-    bundle exec rake
-
-    # Javascript tests.
-    RAILS_ENV=test bundle exec rake spec:javascript
-
+### Ruby
+```bash
```
juliusv: whitespace nit: for making things easier to read (especially in the JSHint section), please put an empty line before and after any code blocks.

We should also be consistent in general with empty lines after headers, though either way is fine, given that we have both variants in this README already.

app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -21,14 +21,15 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
       $scope.widgets = $scope.widgets || [];
       var urlVars = URLVariablesDecoder();
       var templateVarRe = /^var\.(.*)$/;
-      for (var o in urlVars) {
-        var matches = o.match(templateVarRe)
+      var o;
```
juliusv: Just wondering: do you know why it suggests this change? IMO it makes the code less readable because now you need to assume that `o` is used elsewhere than just in the loop.

app/assets/javascripts/angular/services/graph_refresher.js
```diff
@@ -21,9 +21,10 @@ angular.module("Prometheus.services").factory('GraphRefresher',
         cache: false
       }).then(function(payload, status) {
         data = payload.data;
+        var errMsg;
```
juliusv: Same comment applies here. Actually, JavaScript's scoping is such that a variable defined in the `switch` below will be visible also outside of the `switch`. Same for loops. Probably that's why. Meh.

juliusv: Still, for readability I think it's better the other way around. Do you know if that check can be disabled, or maybe there's good arguments for not doing so?

juliusv:
Woot! Seems like about time, given all the little stuff it discovered :)

CC @grobie because he'll like this.

:+1: besides the comments.

app/assets/javascripts/angular/resources/shared_graph_behavior.js
stuartnelson3: javascript scoping rules :/ the o is actually available outside of the `for-in` loop, even in the original declaration `for (var o in urlVars)`

app/assets/javascripts/angular/services/graph_refresher.js
stuartnelson3: it has been turned off

.travis.yml
```diff
@@ -6,8 +6,10 @@ before_script:
   - mysql -e 'create database promdash_test;'
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+  - npm install jshint
 script:
   - cp config/database.yml.example config/database.yml
   - bundle exec rake db:migrate
   - bundle exec rake spec
   - bundle exec rake spec:javascript
+  - jshint app/assets/javascripts
```
grobie: Maybe execute this as the first thing? Some syntax errors cause spec issues, so failing fast for them might improve test run time.

grobie: Also, maybe add a `make test` target to run all specs and jshint?

grobie:
Yay! More time for @juliusv to focus on actual changes :)

PR #324, opened 2015-02-04
Add zoom-out functionality to breakout widgets

stuartnelson3:
Moves the 'z' key listener to the shared graph behavior object, instead
of keeping it isolated in the dashboard_ctrl.


juliusv:
:+1:

PR #325, opened 2015-02-04
Fix frame height calculation

stuartnelson3:
the class being targeted for generating frame height was on the wrong element, causing frames to be slightly larger than graphs.


juliusv:
:+1:

PR #326, opened 2015-02-06
Refresh frame with updated endtime/range values.

stuartnelson3:
The refresh was aborting early in the previous commit, causing changes in just endtime/range not to be added to the url. Now, instead of a transient url var, the scope stores the current escaped url. Updated fields e.g. height width from until will be written to this url, which is used as the source for the iframe.

@juliusv

### DIFF
--- a/app/assets/javascripts/angular/controllers/frame_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/frame_ctrl.js
@@ -92,7 +92,6 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
         return o.key;
       }
     }).join("&");
-    $scope.frame.escapedURL = VariableInterpolator(parser.href, $scope.vars);
     $scope.frame.url = unescape(parser.href);
   }
   initFrameURL();
--- a/app/assets/javascripts/angular/directives/frame.js.erb
+++ b/app/assets/javascripts/angular/directives/frame.js.erb
@@ -17,19 +17,6 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
         return WidgetHeightCalculator(element.find(".js_widget_wrapper")[0], scope.aspectRatio);
       }
 
-      scope.refreshFrame = function() {
-        var frame = element.find('.frame_container');
-        frame.html(frame.html());
-      };
-
-      var a = ['frame.url', 'frame.range', 'frame.endTime', 'refreshCounter', 'frame.graphite'];
-      a.forEach(function(w) {
-        scope.$watch(w, regenFrame);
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -17,10 +17,7 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
         return WidgetHeightCalculator(element.find(".js_widget_wrapper")[0], scope.aspectRatio);
       }
 
-      scope.refreshFrame = function() {
-        var frame = element.find('.frame_container');
-        frame.html(frame.html());
-      };
+      scope.refreshFrame = regenFrame;
```
juliusv: Why not put the function directly on the scope instead of having it as a separate name that's then assigned here?

juliusv:
:+1: besides comment

app/assets/javascripts/angular/controllers/frame_ctrl.js
```diff
@@ -92,7 +92,6 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
         return o.key;
       }
     }).join("&");
-    $scope.frame.escapedURL = VariableInterpolator(parser.href, $scope.vars);
```
stuartnelson3: I got rid of this escapedURL monkey business in favor of only interpolating into the URL when we make the actual iframe.

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -37,17 +24,21 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
...
       // history state added by modifying an iframe's src.
       // http://khaidoan.wikidot.com/iframe-and-browser-history
-      function regenFrame() {
-        if (!scope.frame.escapedURL) {
-          return;
-        }
-        var url = scope.frame.escapedURL.slice();
-        delete scope.frame.escapedURL;
+      scope.refreshFrame = function() {
+        var url = VariableInterpolator(scope.frame.url, scope.vars);
```
stuartnelson3: this makes it much easier.

stuartnelson3:
@juliusv hey updated the code to make it simpler, let me know when you get the chance to take a look

juliusv:
:+1:

grobie:
Missing tests :P

stuartnelson3:
yeah, need additional testing stuff to test this :/ getting a fix out was more important

https://github.com/karma-runner/karma-html2js-preprocessor

grobie:
Cool. Thanks for the update.

PR #328, opened 2015-02-09
Fix: Rails 4 upgrade; added 'bin' into source control.

scharissis:
Hi,

A fresh clone of promdash fails to start a rails server. After a `bundle install` an attempt to run `bundle exec rails server` will throw a usage error of `rails new`. I believe this is because the loader failed to find the appropriate files and thus deemed the root dir was not a rails app dir.
As per the comments in https://github.com/rails/rails/blob/master/railties/lib/rails/app_rails_loader.rb I've 'updated' to rails 4, by including these bin files.
Now I can run a local rails server.

These are my application versions:
$ gem -v
2.2.2
$ bundle -v
Bundler version 1.7.13
$ rails -v
Rails 4.1.7

OS: Ubuntu 14.10

If this isn't the root cause or an acceptable fix, feel free to ignore/reject this PR.

Stefano

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@
 .ruby-gemset
 #.ruby-version
 
-/bin
 /build
 /vendor/bundle
 /system
--- a/bin/bundle
+++ b/bin/bundle
@@ -1,0 +1,3 @@
+#!/usr/bin/env ruby
+ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
+load Gem.bin_path('bundler', 'bundle')
--- a/bin/rails
+++ b/bin/rails
@@ -1,0 +1,4 @@
+#!/usr/bin/env ruby
+APP_PATH = File.expand_path('../../config/application',  __FILE__)
+require_relative '../config/boot'
+require 'rails/commands'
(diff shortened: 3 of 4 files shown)


juliusv:
@scharissis Thanks for sending this in!

@grobie We added the `bin` subdir to `.gitignore` a long while ago, and I think you were involved in that decision back then. What's your take on this whole checking-in-binstubs now? What's the "right" thing to do with Rails 4? In any case, if we don't check them in, we need to change the `README.md` instructions to generate them.

grobie:
I'm not up-to-date with Rails conventions anymore.

Checking in binstubs sounds fine to me, I ran into the same issues.
Alternatively we'd need to document that `bundle exec rake rails:update:bin` is a necessary setup step. That could even be part of a `make setup` target.

juliusv:
Cool. Actually, the linked comments in https://github.com/rails/rails/blob/master/railties/lib/rails/app_rails_loader.rb seem to be clear. So this looks good :+1: Thanks @scharissis, merging.

juliusv:
@scharissis Oh, and feel free to add yourself to the AUTHORS file.

scharissis:
No problem!
Cool, I updated the docs a bit too and did this in a new PR: https://github.com/prometheus/promdash/pull/329

juliusv:
Woot, thanks a bunch!

PR #331, opened 2015-02-12
Update base packages and cleanup apt cache

sammcj:
I noticed when building this that it pulled in an insecure version of OpenSSL from ubuntu:13.10 - It's important to make sure that such packages are kept up to date so adding an apt-get upgrade at the start of the dockerfile will prevent this.

An additional suggestion would be to switch to a Debian base image rather than Ubuntu - Debian's packages are better maintained and generally have faster security patching - for newer packages you can also include Debian backports sources.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM       ubuntu:13.10
 MAINTAINER Prometheus Team <prometheus-developers@googlegroups.com>
 ENV        RAILS_ENV production
 
-RUN        apt-get update && apt-get install -yq \
-           ruby2.0 ruby2.0-dev gcc make g++ libmysqlclient-dev
+RUN        apt-get update && apt-get -qq upgrade && apt-get install -yq \
+           ruby2.0 ruby2.0-dev gcc make g++ libmysqlclient-dev && apt-get clean
 RUN        gem install bundler
 
 WORKDIR    /promdash


juliusv:
@discordianfish

discordianfish:
@sammcj Good catch, openssl is indeed vulnerable in ubuntu:13.10. I'll foward that report and will make sure the image gets updated.

But still, we shouldn't run 'upgrade' on the image because it will increase the image size a lot. But I agree, we should switch to debian. Would you mind submitting a PR to use debian instead?

discordianfish:
@sammcj Actually, if possible we might want to use the official 'ruby' image.

discordianfish:
13.10 is EOL, for now I created #338

sammcj:
@discordianfish that's a much better idea - https://registry.hub.docker.com/_/ruby/ looks pretty well thought out.

Using the `2.2.0-wheezy` tag will get you Ruby 2.2.0 with a Debian Wheezy base (current stable).

While running an upgrade on a higher layer does increase the image size (which is a pain), it also means that you're ensuring the application is secure - so I guess it's a bit of a trade-off - perhaps the answer is to allow users to decide for themselves in some easy way?

discordianfish:
@sammcj Even if you do a apt-get upgrade in the prometheus image, we would still need to rebuild it regularly. The Docker Hub's automated build feature pulls in the latest base image on every build anyways, so the packages will be up to date after a build anyways. Of course, given that the base images are up to date which is at least usually the case :)

juliusv:
@discordianfish @sammcj Should this PR be closed?

Issue #332, opened 2015-02-12
PostgreSQL support - promdash assumes people will use MySQL


sammcj:
Promdash makes the assumption that people want to use, and only installs the gem for MySQL.
It would be great it PostgreSQL supporting gem could also be added by default.
https://github.com/prometheus/promdash/blob/master/Gemfile#L6

juliusv:
Fixed in #334. Thanks!

PR #333, opened 2015-02-12
Add missing gems for PostgreSQL support

sammcj:
As per https://github.com/prometheus/promdash/issues/332

I'm actually a little unsure if activerecord-postgresql-adapter is truly required, or if pg is enough?


sammcj:
Cancel this, I'm going to re-do

PR #334, opened 2015-02-12
Add missing gems for PostgreSQL support

sammcj:
Also removed Gemfile.lock and added to .gitignore as that probably shouldn't have been committed to the repo?

Fixes: https://github.com/prometheus/promdash/issues/332

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,9 @@ source 'https://rubygems.org'
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '4.1.7'
 
-gem 'mysql2'
+# Allow people to ignore the database provider they don't need by doing bundle install --without mysql
+gem 'mysql2', group: 'mysql'
+gem 'pg', group: 'postgresql'
 
 gem 'active_model_serializers'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,6 +84,7 @@ GEM
     mysql2 (0.3.13)
     nokogiri (1.6.3.1)
       mini_portile (= 0.6.0)
+    pg (0.18.1)
     phantomjs (1.9.8.0)
     pry (0.9.12.2)
       coderay (~> 1.0.5)
@@ -197,6 +198,7 @@ DEPENDENCIES
   json-schema
   launchy
   mysql2
+  pg
   pry
   rails (= 4.1.7)
   rake


sammcj:
Hmm build failures look unrelated?

grobie:
Thanks for the contribution. I think we should keep the `Gemfile.lock` in version control. The app is only tested with the locked dependencies. The behavior the newest possible gem versions is unspecified.

While I have to admit I'm not an active ruby developer anymore, I assume this explanation is still valid: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

sammcj:
@grobie - Just had a chat to a co-worker about this actually and he agreed that it probably is best to keep the .lock in version control.

The problem is when you want to add things to it, it needs to be re-generated - or are you happy to have just the relevant lines added?

For example to add postgresql support:

```diff
87d86
<     pg (0.18.1)
201d199
<   pg
```

to match the addition of 

```
gem 'pg'
```

In the Gemfile

- If you're happy with that, I'll re-add the Gemfile.lock with the included requirements for pg to the PR.

grobie:
Yes, every change to the Gemfile should also be reflected in an updated Gemfile.lock. :+1: to that.

I wonder how to deal with the "different database gems" issue in general. Do you know how other apps handle that? It seems funny to force postgresql people to install the mysql gem and vice versa. Would it be better to not check in any of them?

sammcj:
I'm far from a developer but I'm assuming it could be a case for something like a Gemfile.local - where you can add any variable gems such as database connectors, middleware (such as statds collectors, newrelic etc...)

sammcj:
Updated Gemfile, Gemfile.lock and squashed commits

Gemfile
```diff
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 gem 'rails', '4.1.7'
 
 gem 'mysql2'
+gem 'pg'
```
grobie: After reading a bit about the topic, I'd suggest to define groups for the two gems.

```ruby
gem 'mysql2', group: 'mysql'
gem 'pg', group: 'postgresql'
```

This will allow people to ignore the database provider they don't need by doing `bundle install --without mysql`.

grobie:
Thanks! I added one last comment. :)

sammcj:
Done - Thanks for the help!

grobie:
Thank you!

diurnalist:
pg gem install fails on Mac OS X Mavericks (with postgresql installed via Homebrew): https://gist.github.com/diurnalist/68bd951aa267736bb969

stuartnelson3:
http://stackoverflow.com/questions/19262312/installing-pg-gem-failure-to-build-native-extension/19620569#19620569

juliusv:
@diurnalist Or do:

    bundle install --without="postgresql [...]"

PR #335, opened 2015-02-12
Update makefile and include postgresql during build

grobie:
The Makefile is a hard dependency of SoundCloud's usage so far. This
change should ensure that it still works after the recent postgresql
addition. A later change should probably allow people to use the
Make targets also when using postgresql.

@stuartnelson3 Please test the build with this branch and merge if successful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BUILDDIR=build
 RUBY_VERSION=$(shell cat .ruby-version | perl -ne "chomp and print")
 RUBY_TAR=ruby-$(RUBY_VERSION).tar.gz
 RUBY_SRC=$(BUILDDIR)/ruby-$(RUBY_VERSION)
-RUBY_URL=ftp://ftp.ruby-lang.org/pub/ruby/2.0/$(RUBY_TAR)
+RUBY_URL?=ftp://ftp.ruby-lang.org/pub/ruby/2.0/$(RUBY_TAR)
 RUBY_BIN=$(DESTDIR)/bin/ruby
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,11 +4,19 @@ BUILDDIR=build
 RUBY_VERSION=$(shell cat .ruby-version | perl -ne "chomp and print")
 RUBY_TAR=ruby-$(RUBY_VERSION).tar.gz
 RUBY_SRC=$(BUILDDIR)/ruby-$(RUBY_VERSION)
-RUBY_URL=ftp://ftp.ruby-lang.org/pub/ruby/2.0/$(RUBY_TAR)
+RUBY_URL=http://files/ruby/$(RUBY_TAR)
```
stuartnelson3: build was failing because ruby-lang.org was not resolving

stuartnelson3:
build works, but verify that my change to `$RUBY_URL` is acceptable

Makefile
juliusv: This doesn't make sense for a public project though... we should figure out why it's not resolving instead? Or we can do `RUBY_URL ?= ...`, where the `?=` allows us to override the env var from outside.

stuartnelson3: this is a public project, but the makefile is sc specific. quoting from grobie:

> The Makefile is a hard dependency of SoundCloud's usage

juliusv: See IRC...

juliusv:
:+1:

grobie:
:+1: to your addition :)

juliusv:
Tested on `promdash-devel` app - works!

PR #337, opened 2015-02-12
Encode frame query parameters properly

diurnalist:
The existing query parameter encoding was OK, but failed for a critical use-case:
graphite URLs containing colors that looked like #00AA00. The # character was causing
the URL to be truncated, as it has a special meaning in a URL.

The fix was not simple, because we display the unescaped URL to users (why?).
We are now manually parsing the query string instead of relying on shoving it into
an anchor tag. For our purposes, it should be fine; promdash doesn't do much w/ this.

### DIFF
--- a/app/assets/javascripts/angular/directives/frame.js.erb
+++ b/app/assets/javascripts/angular/directives/frame.js.erb
@@ -1,4 +1,13 @@
-angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "VariableInterpolator", "GraphiteTimeConverter", "WidgetHeightCalculator", function($sce, VariableInterpolator, GraphiteTimeConverter, WidgetHeightCalculator) {
+angular.module("Prometheus.directives").directive('inlineFrame', ["$sce",
+                                                  "VariableInterpolator",
+                                                  "GraphiteTimeConverter",
+                                                  "WidgetHeightCalculator",
+                                                  "URLParser",
+                                                  function($sce,
+                                                    VariableInterpolator,
+                                                    GraphiteTimeConverter,
+                                                    WidgetHeightCalculator,
+                                                    URLParser) {
   var FRAME_CONTAINER_PADDING = 10;
   var FRAME_BACKGROUND_COLOR = '%23191919';
   return {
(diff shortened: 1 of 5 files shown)


diurnalist:
So, I couldn't get the test suite working locally. RAILS_ENV=test had issues connecting to mysql, RAILS_ENV=development would run 0 tests if I ran the javascript rake target... Not sure what the issue is.

stuartnelson3:
`bundle exec rake` should run your tests. you need mysql running on your machine, which I believe you can start with `$ mysql.server start` if you have it installed with `brew`.

`$ RAILS_ENV=test bundle exec rake spec:javascript` should be all you need to run the javascript tests.

we display unescaped urls to users because it was requested that way: https://github.com/prometheus/promdash/issues/280

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -41,20 +41,27 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
       });
 
       function buildFrameURL(url) {
-        var parser = document.createElement('a');
-        parser.href = url;
+        var parts = url.split('?'),
+            path = parts[0],
+            query = parts[1].split('&');
```
stuartnelson3: we prefer to declare `var foo;` on each line

stuartnelson3: also, `query = parts[1].split('&')` throws an error if there is no query string (see `web-status`, e.g.)

diurnalist:
@stuartnelson3 OK, handled those two issues. Adding a test for the directive ended up being very annoying. The html template file is requested over the wire, and mocking that is pretty gross in jasmine. If karma was integrated in to the project, there is a path forward with [karma-ng-html2js-preprocessor][1].

[1]: https://github.com/karma-runner/karma-ng-html2js-preprocessor

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -41,20 +41,27 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
       });
 
       function buildFrameURL(url) {
-        var parser = document.createElement('a');
-        parser.href = url;
+        var parts = url.split('?'),
```
stuartnelson3: errant `,` here is breaking things

stuartnelson3:
yeah, the preprocessor definitely needs to be added. all the most complicated logic exists in the directives.

app/assets/javascripts/angular/directives/frame.js.erb
diurnalist: :ultimatefacepalm:

stuartnelson3: :P

app/assets/javascripts/angular/directives/frame.js.erb
```diff
@@ -66,13 +73,14 @@ angular.module("Prometheus.directives").directive('inlineFrame', ["$sce", "Varia
...
-        parser.search = '?' + decodeURI($.param(fields)) + '&' + targets.join('&') + '&decache=' + scope.refreshCounter;
-        return parser.href;
+
+        return [
+          'height=' + height,
+          'width=' + ((height / scope.aspectRatio) - FRAME_CONTAINER_PADDING),
+          'from=' + GraphiteTimeConverter.graphiteFrom(scope.frame.range, scope.frame.endTime),
+          'until=' + GraphiteTimeConverter.graphiteUntil(scope.frame.endTime),
+          'bgcolor=' + FRAME_BACKGROUND_COLOR
+        ].concat(targets);
```
grobie: Doesn't this change the behavior? There are more allowed fields in graphite than targets and the five defined above. We don't want to remove them. Can you explain the motivation of the change and how other fields like `ymax=500` will be preserved?

diurnalist:
@grobie @stuartnelson3 sorry, didn't have time to look at this until today. Fixed up things so it's more of a real solution. Added a URLParser library that will work for promdash's needs. Unfortunately couldn't add a spec for the directive b/c of the `asset_path` rails thing - we'd probably need an additional karma preprocessor to find those and strip out that ERB stuff.

diurnalist:
@stuartnelson3 nice job on getting karma in there, btw - much easier :)

grobie:
This looks great! Thank you so much @diurnalist. :+1:, but I'll leave the actual js review to @stuartnelson3 ;)

stuartnelson3:
looks good, :+1: !

PR #338, opened 2015-02-12
Use ubuntu 14.04 instead of EOL 13.10

discordianfish:
Do not merge yet, still need to make sure it works.


juliusv:
:+1: thanks!

discordianfish:
Read the message ;) But no worries, will fix that tomorrow - until then, the old version is still available.

Issue #341, opened 2015-02-13
Allow for servers with basic auth


oliver006:
I have a prometheus server sitting behind nginx and use basic auth to restrict access.
Currently, promdash fails when I try to add datasources from that server.
It'd be great if basic auth would work.

juliusv:
Just a thought: does it work to encode the username/password into your server URL that you configure in PromDash? I.e. as the Prometheus server URL that you configure in PromDash, provide `http://<user>:<pass>@1.2.3.4:9090/`

ericflo:
I just tried that on my install, and it didn't seem to work.

oliver006:
Yeah, tried that but it didn't work.
I checked the browser console, it's doing a straight request to the url without the basic auth credentials which results in the 401.

juliusv:
Indeed, I can confirm that. But in general, configuring username/password pairs for each server in PromDash and storing the password plaintext in the PromDash database would be ok? (I don't really see a way around that?)

juliusv:
So I've been playing with AngularJS/CORS/BasicAuth/nginx locally for a few hours now and haven't been able to tweak the settings on all parts of the chain in such a way that the combination works. I suspect that nginx's basic auth `401` is messing up the CORS flow somehow, which in turn prevents sending any Authorization headers at all... but not sure. If anyone wants to experiment with this or take this on, I'd be greatful.

oliver006:
Thanks for looking into this!
I played around with the nginx side of things and I got it to work by adding an nginx location for `/api` and make that proxy-pass through to my prometheus server at `localhost:9090` now I cna use the same servername as the dashboard for the prometheus server and auth and CORS both work.
Once #343 lands it'd much easier and cleaner to have different locations proxy to different prometheus servers but this works for me for now.

juliusv:
@oliver006 Oh interesting. Did you modify PromDash to add basic auth? Also, are you actually doing CORS or avoiding it by having both PromDash and Prometheus going through the same server name (via the proxy)? Would be interesting to hear more about your solution.

#343 becomes most useful together with supporting different roots in the Prometheus server itself, but that PR (https://github.com/prometheus/prometheus/pull/528) is currently stuck because it would break some other things like having console templates in the server at arbitrary directory depths (relative asset paths would then change depending on directory depth). But actually, that latter PR should only be relevant for using the Prometheus web UI, not the API. If all you want to do is use the API from PromDash, that should work with #343 alone.

oliver006:
[...]

PR #343, opened 2015-02-14
Allow server URLs with custom paths or no trailing slash.

juliusv:
This enables people to use PromDash behind a proxy with a custom path,
like `http://myproxy/prometheus`.

### DIFF
--- a/spec/javascripts/angular/services/url_generator_spec.js
+++ b/spec/javascripts/angular/services/url_generator_spec.js
@@ -1,0 +1,27 @@
+//= require spec_helper
+describe('URLGenerator', function() {
+  var urlGenerator;
+  beforeEach(inject(function(_URLGenerator_) {
+    urlGenerator = _URLGenerator_;
+  }));
+
+  it('allows urls with custom paths, no trailing slash', function() {
+    ['http://promdash.server.com/prometheus', 'http://promdash.com'].forEach(function(s) {
+      ['/api/query_range', '/api/query', '/api/metrics', '/arbitrary/endpoint'].forEach(function(ep) {
+        var s = 'http://promdash.server.com/prometheus';
+        var url = urlGenerator(s, ep);
+        expect(url).toEqual(s + ep);
+      });
+    });
+  });
+
+  it('allows urls with custom paths, with trailing slash', function() {
+    ['http://promdash.server.com/prometheus/', 'http://promdash.com/'].forEach(function(s) {
+      ['/api/query_range', '/api/query', '/api/metrics', '/arbitrary/endpoint'].forEach(function(ep) {
+        var s = 'http://promdash.server.com/prometheus/';
+        var url = urlGenerator(s, ep);
+        expect(url).toEqual(s.substring(0, s.length - 1) + ep);
+      });
+    });
+  });
+});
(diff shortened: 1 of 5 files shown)


juliusv:
@stuartnelson3

juliusv:
When looking at adding tests, I was going back and forth between either adding three tests (one for each controller/service), which would also require me to figure out how to test successful graph/pie/metric-name loads, or extracting the URL concatentation code into a separate service and testing that once. Both seemed like a lot of overhead for such a simple orthogonal change though. Let me know if you think otherwise...

grobie:
It must not be hard to add tests or feel like it's an overhead. That means
we need more work on the test suite. Almost all changes are itself simple,
the combination of all these simple things make a program eventually
complex and is the reason we need a test suite.

I vote to extract all the HTTP request handling here and have a common
service for that. Regardless of tests, doing the same change N times (3
right now, but will we remember this when the fourth endpoint is added?)
doesn't seem to be right. For #341 it's also necessary to set the
Authentication header correctly for prometheus server requests
https://docs.angularjs.org/api/ng/service/$http.

juliusv:
Agreed, together with the authentication header change for #341 it would make much more sense to split this out into a separate HTTP service, and then test that. I'm currently struggling to get the combination of Angular/CORS/Basic Auth working... it's a pain. The headers are just not set, etc.

juliusv:
(oh yeah, and with nginx as the basic auth proxy in the middle)

stuartnelson3:
I'll add on to what @grobie said: it being hard to write a test is usually indicative of writing our code incorrectly. obviously we don't have enough hours in the day to fix everything, but I do intend on spending more time on the test suite.

also, i'll review this tomorrow :P

stuartnelson3:
@juliusv added tests

juliusv:
Oh, thanks so much! :) Adding some tiny comments now...

juliusv:
Actually, just :+1: 

Once we do end up adding basic auth, we probably want to not only extract the URL building, but the whole HTTP call. But we can leave that for then.

spec/javascripts/angular/services/url_generator_spec.js
```diff
@@ -0,0 +1,27 @@
...
+describe('URLGenerator', function() {
+  var urlGenerator;
+  beforeEach(inject(function(_URLGenerator_) {
+    urlGenerator = _URLGenerator_;
+  }));
+
+  it('allows urls with custom paths, no trailing slash', function() {
+    ['http://promdash.server.com/prometheus', 'http://promdash.com'].forEach(function(s) {
+      ['/api/query_range', '/api/query', '/api/metrics', '/arbitrary/endpoint'].forEach(function(ep) {
+        var s = 'http://promdash.server.com/prometheus';
```
grobie: I'm confused. First you have a `s` variable which can contain two different domain names, but then you override `s` to this server name only? Or what am I missing here?

juliusv: Oh, good catch. Yeah, looks like this line (and the same one in the test below) should be deleted?

stuartnelson3: ahhh this is true. when I expanded the test to have the two different server names, i forgot to delete the original `s` var.

PR #344, opened 2015-02-16
Allow range and until vars to propagate into widget on single widget

stuartnelson3:
@juliusv


juliusv:
:+1:

I'm a bit wary of the general architecture of how the shared graph behavior reaches "into" the widgets and does stuff to them conditionally depending on their type, instead of the widgets receiving some signal (broadcast?) and doing the right thing themselves. Missing encapsulation of implementation details, that is. That's a bit of a meta-topic though, so :+1: for this.

PR #354, opened 2015-02-24
add ignoreHash option to include all colors in colorList

stuartnelson3:
@diurnalist see the test added. the issue being seen was that users would have a list of colors, and the final hashtag was being set as `hashIndex` in the constructor.

instead of `http://graphite-example.com/render?colorList=#38AFEB,#EB3838,#38EB38,#EB38DC&decache=3`, the url was being stringified to
`http://graphite-example.com/render?colorList=#38AFEB,#EB3838,#38EB38,&decache=3#EB38DC` (and breaking)

Is this how I should handle this issue?

### DIFF
--- a/app/assets/javascripts/angular/controllers/frame_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/frame_ctrl.js
@@ -58,7 +58,7 @@ angular.module("Prometheus.controllers").controller('FrameCtrl', ["$scope",
 
   $scope.generateFrameComponents = function() {
     // This URL should already be unescaped
-    var parser = URLParser($scope.frame.url);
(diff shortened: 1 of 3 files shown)


stuartnelson3:
@diurnalist ping

diurnalist:
Yes, that will solve this problem. This used to "work" by looking at the "this is a graphite url" flag. If that was set, I believe it would use the ignoreHash behavior. Otherwise it would treat that character as denoting the beginning of the hash part of the URL.

This change sort of means that hashes are not supported in promdash URLs, unless I'm missing something?

stuartnelson3:
yeah that was my concern. so I'll change it so that "this is a graphite url" means to ignore hash, which seems like the right solution

diurnalist:
:fist:

PR #358, opened 2015-02-27
Prevent widget overflow

grobie:
promdash - overflow
promdash - overflow hidden


stuartnelson3:
this hides graph datapoint tooltips
image
image

grobie:
Hm. I see. I guess there is no way around it? Closing for now until you have an idea.

Issue #352, opened 2015-02-20
Auto-refresh when changing servers


grobie:
If a different server gets selected, the graph should be auto-refreshed immediately, like it happens when changing time.

juliusv:
This is actually not specific to the server selection - we currently don't refresh the data automatically for any of the per-graph inputs (including time). I think it should refresh automatically for:

* changing time settings
* changing the resolution
* changing the server
* when pressing enter in the expression input (but not when just changing it, I guess - it would be too noisy)

matthiasr:
Seconded! It's annoyingly hard to get a new expression to evaluate.

grobie:
Thanks!

PR #360, opened 2015-03-04
Refresh graph according to situations enumerated in #352

stuartnelson3:
requested in #352

### DIFF
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -102,6 +102,10 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
     $scope.refreshGraph(500);
   });
 
+  $scope.$on('changeExpression', function(ev) {
+    $scope.refreshGraph();
+  });
+
   $scope.addLegendString = function() {
     var lsts = $scope.graph.legendFormatStrings;
     var id = (new Date()).getTime().toString(16);
@@ -143,7 +147,18 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
     $scope.addAxis();
   }
 
+  $scope.$watch('graph.range', function() {
+    $scope.refreshGraph();
+  });
+  $scope.$watch('graph.endTime', function() {
+    $scope.refreshGraph();
+  });
+  $scope.$watch('graph.resolution', function() {
+    $scope.refreshGraph();
+  });
+
   $scope.refreshGraph();
+
   if (location.pathname.match(/^\/w\//)) { // On a widget page.
     $scope.widgetPage = true;
     $scope.dashboard = dashboard;
--- a/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
@@ -155,7 +155,7 @@ angular.module("Prometheus.controllers")
       title: "Title",
       expression: {
         id: 0,
-        serverID: 1,
+        serverID: ($scope.servers[0] || {}).id,
         expression: "",
         legendID: 1
       },
--- a/app/assets/javascripts/angular/controllers/pie_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/pie_ctrl.js
@@ -13,7 +13,7 @@ angular.module("Prometheus.controllers").controller('PieCtrl',
   $scope.refreshGraph = function() {
     var exp = $scope.graph.expression;
     var server = $scope.serversById[exp.serverID || 1];
-    if (server === undefined) {
+    if (server === undefined || !exp) {
       return;
     }
     $scope.requestInFlight = true;
@@ -46,6 +46,10 @@ angular.module("Prometheus.controllers").controller('PieCtrl',
   };
 
   $scope.refreshGraph();
+
+  $scope.$on('changeExpression', function(ev) {
+    $scope.refreshGraph();
+  });
 
   if (location.pathname.match(/^\/w\//)) { // On a widget page.
     $scope.widgetPage = true;
--- a/app/assets/javascripts/angular/controllers/pie_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/pie_ctrl.js
@@ -13,7 +13,7 @@ angular.module("Prometheus.controllers").controller('PieCtrl',
   $scope.refreshGraph = function() {
     var exp = $scope.graph.expression;
     var server = $scope.serversById[exp.serverID || 1];
-    if (server === undefined || !exp) {
+    if (server === undefined || !exp.expression) {
       return;
     }
     $scope.requestInFlight = true;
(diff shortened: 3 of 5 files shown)


stuartnelson3:
@juliusv @grobie poke

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -143,7 +147,18 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
     $scope.addAxis();
   }
 
+  $scope.$watch('graph.range', $scope.refreshGraph);
+  $scope.$watch('graph.endTime', $scope.refreshGraph);
```
juliusv: This doesn't seem to fire for the end time, neither when clicking on the next/prev buttons, nor when selecting a time via the datapicker. It does update when I click the "x" clear button though. Can you reproduce?

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -143,7 +147,18 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
...
+  $scope.$watch('graph.resolution', $scope.refreshGraph);
+  $scope.graph.expressions.forEach(function(e) {
```
juliusv: This only covers expressions that are there at dashboard load, but not new ones... I guess the expression would have to fire a `setServer` event (or similar), like you do with the changed expression?

app/assets/javascripts/angular/controllers/graph_ctrl.js
stuartnelson3: ah, that is truly spectacular.

the arguments to the callback function are `function(newValue, oldValue)`.
in this case `$scope.refreshGraph` was the function, and `newValue` could be an enormous integer representing the unix time for the endtime. this was interpreted as the debounce time for `$scope.refreshGraph`, hence looooong waits for the refresh to occur. updated.

app/assets/javascripts/angular/controllers/graph_ctrl.js
stuartnelson3: Worked around this by making a more explicit watch function. Changes pushed.

stuartnelson3:
updated

grobie:
If I understand the PR correctly, this will refresh the graph when pressing enter in an expression input field. Can we also trigger a refresh when the input field loses it's focus?

stuartnelson3:
Do we want to make that assumption? I wouldn't expect that behavior, personally.

juliusv:
Me neither... if it loses focus, I might not even be finished typing my
expression, but copy/pasting stuff from somewhere else, etc.

grobie:
What downsides do you see? Why wouldn't you expect it?

If I edit an expression I expect the graph to reflect the changes I made.
As refreshing after every change is too noisy and error-prone, we need
clear signals when to trigger a refresh. Pressing ENTER is one of them,
leaving the field seems to be the most common one?

The refresh should probably only be triggered if the input field content
has actually changed.

grobie:
I see, fair enough. I'll train to hit enter then.

juliusv:
Hm, btw. what about pie charts?

stuartnelson3:
promdash has pie charts?

juliusv:
Ok, one last thing: when adding a new pie chart, it has one (empty) expression by default. That is good (and the graph should actually be changed to also have one initial expression after creating), but we should ignore empty expressions during graph refreshes, as they'll otherwise generate errors. Right now, when you create a pie chart, you immediately get an error...

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -143,7 +147,32 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
...
+  $scope.$watch('graph.endTime', function() {
+    $scope.refreshGraph();
+  });
+  $scope.$watch('graph.resolution', function() {
+    $scope.refreshGraph();
+  });
+
+  $scope.$watch(function() {
+    return $scope.graph.expressions.filter(function(e) {
+      // Filters newly added expressions from $watch, so it
```
juliusv: Wouldn't it be simpler and more consistent to also just emit a `setExpression` (probably then renamed to `changeExpression` to cover both changes to the text and the server) event for changes to the server ID? Then the code here can disappear and you'd only have to bind a change handler in the `expression.js.erb`. Together with ignoring empty expressions in general, that should be fine, or am I missing something?

stuartnelson3: :+1:

stuartnelson3:
updated

juliusv:
Hm, but the problem with the pie chart empty expression error (and empty expression errors in general remains)?

stuartnelson3:
What is the problem? Do you not want the user to be able to make a request if the input is empty? 

I thought the issue was that when changing servers it would force a refresh, which no longer occurs if the expression box is empty

juliusv:
The problem is:

- create a new pie chart
- you get an error `Error parsing rules at line 1, char 1: syntax error` because the initial expression is empty

So I think empty expressions should always be ignored in any case, yes. Then I meant relatedly (but optional for this PR) that it would be great to also add an initial (empty) expression to a new graph, as that wouldn't trigger an error anymore if we ignore empty expressions for queries.

stuartnelson3:
when do you get the error? i create a new pie chart, select a server, and get no error. i only get an error when i force a refresh. grab me when you're free.

juliusv:
I get the error immediately when clicking the "Add Pie Chart" button with no further interaction required... unfortunately I'm WFH today because a maintenance person needs access.

stuartnelson3:
ah, i do not get that. try clearing your cache, repulling the branch, etc, etc...

juliusv:
Did both, including hard-refresh in incognito mode :)

```
$ git rev-parse HEAD                  
668a1d090203e063c2b0f4c3492c2a8f7ca5c5ec
```

My PromDash has a single server configured, and if I create a completely new dashboard and immediately click "Add pie chart", I get that error. Anyways, if you can't reproduce it, we can take a look at it tomorrow, when I'm back in the office.

stuartnelson3:
yeah, just followed your steps and can't reproduce it. catch up with you tomorrow.

juliusv:
@stuartnelson3 Funny, it doesn't happen on my work laptop either. Guess you'll need to come to my place finally to see the problem on my desktop one :) I'll check again at home to see what the difference is.

stuartnelson3:
computers, amirite?

juliusv:
Found the reason: https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/controllers/dashboard_ctrl.js#L158

When you have a server with ID 1, it will be selected by default and you'll get an error. That shouldn't happen (either don't select any server by default or always ignore empty expressions and preselect the first available server - the latter is the best solution IMO).

stuartnelson3:
ah k.

the first server should be selected by default now, and `changeExpression` will only be emitted if the pie chart has an expression to evaluate.

juliusv:
Hmm, I'm still getting an error about the empty expression when adding a new pie chart. Stop by my desk in case you can't reproduce it...

stuartnelson3:
finally got the error, no idea what was wrong with my machine. when you get a chance let me know if this *finally* fixes it

juliusv:
@stuartnelson3 Hum, nope, still getting the error. Wanna come over and take a look?

stuartnelson3:
@juliusv copacetic?

juliusv:
Yup :+1:

PR #364, opened 2015-03-23
Remove unused decache query param

stuartnelson3:
Removes `$scope.refreshCounter`, which was no longer being used.

@juliusv


juliusv:
:+1:

PR #365, opened 2015-03-27
promdash gauge widget

stuartnelson3:
gauge widget

**todo**
- [x] add `gauge` to json schema definition
- [x] re-enable schema validation on dashboard save
- [x] add option to only display value (no gauge)

### DIFF
--- a/app/assets/templates/gauge_template.html.erb
+++ b/app/assets/templates/gauge_template.html.erb
@@ -1,0 +1,96 @@
+<span ng-controller="GaugeCtrl">
+  <div class="widget_wrapper panel panel-default pull-left">
+    <error-message error-messages="errorMessages"></error-message>
+    <div class="js_widget_wrapper widget_title panel-heading" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
+      <span ng-show="dashboard"><a href="/{{dashboard.slug}}" class="dashboard_link pointer">{{dashboard.name}}</a> &mdash; </span>
+      {{title()}}
+
+      <div class="btn-group pull-right graph_config_menu">
+        <div ng-show="showControls || showTab || requestsInFlight" class="btn btn-primary btn-sm" ng-click="refreshGraph()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestInFlight}" class="glyphicon glyphicon-refresh"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'datasources')" tooltip-append-to-body="true" tooltip="Datasources"><i class="icon-add-to-list"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'graph')" tooltip-append-to-body="true" tooltip="Chart settings"><i class="icon-gauge"></i></div>
+        <div ng-show="(showControls || showTab) && !widgetPage" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'staticlink'); generateWidgetLink($event)" tooltip-append-to-body="true" tooltip="Link to Chart"><i class="icon-link"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="widgetDeleteModal()" tooltip-append-to-body="true" tooltip="Remove chart"><i class="icon-trash"></i></div>
+      </div>
+
+    </div>
+
+    <div class="pull-left graph_control_tabbar" ng-show="showTab">
+      <!-- Chart control tab content panes -->
+      <!-- 1) Datasources -->
+      <div class="graph_control_tabpane panel panel-default" ng-show="showTab == 'datasources'">
+        <div class="panel-heading">
+          Datasources
+          <i ng-click="closeTab($event)" class="clone_widget_close pointer pull-right icon-cross"
+           tooltip-append-to-body="true" tooltip="Close"></i>
+        </div>
+        <div class="panel-body">
+          <expression
+            axes="graph.axes"
+            end-time="graph.endTime"
+            expr="graph.expression"
+            range="graph.range"
+            servers="servers"
+            servers-by-id="serversById"
+            type="{{graph.type}}"
+            vars="vars"
+            index="$index"></expression>
+        </div>
+      </div>
+
+      <!-- 2) Chart Settings -->
+      <div class="graph_control_tabpane panel panel-default" ng-show="showTab == 'graph'">
+        <div class="panel-heading">
+          Chart Settings
+          <i ng-click="closeTab($event)" class="clone_widget_close pointer pull-right icon-cross"
+           tooltip-append-to-body="true" tooltip="Close"></i>
+        </div>
+        <div class="panel-body">
+          <label>Gauge Title:</label>
+          <input class="form-control" ng-model="graph.title">
+          <label>Gauge Max:</label>
+          <input class="form-control" type="number" ng-model="graph.max">
+          <label>Gauge Units:</label>
+          <input class="form-control" ng-model="graph.units">
+          <label>Refresh interval (seconds):</label>
+          <input class="form-control" type="number" ng-model="graph.refresh" min="1">
+          <label>Numeric display precision (decimal points):</label>
+          <input class="form-control" type="number" ng-model="graph.precision" min="0">
+          <label>Warning threshold (percent):</label>
+          <input class="form-control" type="number" ng-model="graph.warning" max="100" min="1">
+          <label>Danger threshold (percent):</label>
+          <input class="form-control" type="number" ng-model="graph.danger" max="100" min="1">
+          <div class="checkbox">
+            <label><input type="checkbox" ng-model="graph.hideGauge" /> Hide graphical gauge</label>
+          </div>
+        </div>
+      </div>
+
+      <!-- 5) Static link to graph -->
+      <div class="graph_control_tabpane panel panel-default" ng-show="showTab == 'staticlink'">
+        <div class="panel-heading">
+          Link to this graph
+          <i ng-click="closeTab($event)" class="clone_widget_close pointer pull-right icon-cross"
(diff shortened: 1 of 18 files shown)


stuartnelson3:
@juliusv can you help me write the schema for the gauge type?

juliusv:
@stuartnelson3 Yup, I'll come over.

dashboard_schema.json
```diff
@@ -312,6 +312,71 @@
               }
             }
           },
+          // "gauge" widget type.
+          {
```
stuartnelson3: @juliusv need your help on this, when i try to save the validation error being returned says only how the gauge doesn't conform to chart/pie/frame, it makes no mention of it not conforming to gauge.

stuartnelson3:
@juliusv tests should pass now, had to update fixture json. ready for your careful eye :P

juliusv:
Getting to this now...

app/assets/templates/gauge_template.html.erb
```diff
@@ -0,0 +1,107 @@
...
+           tooltip-append-to-body="true" tooltip="Close"></i>
+        </div>
+        <div class="panel-body">
+          <label>Gauge Title:</label>
+          <input class="form-control" ng-model="graph.title">
+          <label>Gauge Max:</label>
+          <input class="form-control" type="number" ng-model="graph.max">
+          <label>Gauge Units:</label>
+          <input class="form-control" ng-model="graph.units">
+          <label>Refresh interval (seconds):</label>
```
juliusv: Can we default this to something saner than `1s`? It'll blow up people's servers otherwise. Maybe default it to the current dashboard refresh interval? Or if that's not readily accessible in this part of the code, maybe just `1m`?

stuartnelson3: sure, whichever you prefer

app/assets/templates/gauge_template.html.erb
```diff
@@ -0,0 +1,107 @@
...
+          <input class="form-control" type="number" ng-model="graph.refresh" min="1">
+          <label>Precision:</label>
```
juliusv: This is a bit vague. Can we call it `Numeric display precision (decimal points)` or something like that?

app/assets/templates/gauge_template.html.erb
juliusv: Let's just go with `1m` unless you have the global refresh rate readily available. But maybe the latter doesn't make sense anyways, since we probably also don't want to update the widget's refresh rate when changing the global refresh.

juliusv: ACTUALLY

juliusv: Can we use the global refresh interval if this is left empty?

juliusv:
Hmm, are you able to save? I'm getting `The property '#/widgets/1' contains additional properties ["range"] outside of the schema when none are allowed`

app/assets/templates/gauge_template.html.erb
```diff
@@ -0,0 +1,107 @@
...
+      <span ng-show="dashboard"><a href="/{{dashboard.slug}}" class="dashboard_link pointer">{{dashboard.name}}</a> &mdash; </span>
+      {{title()}}
+
+      <div class="btn-group pull-right graph_config_menu">
+        <div ng-show="showControls || showTab || requestsInFlight" class="btn btn-primary btn-sm" ng-click="refreshGraph()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestInFlight}" class="glyphicon glyphicon-refresh"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'datasources')" tooltip-append-to-body="true" tooltip="Datasources"><i class="icon-add-to-list"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'graph')" tooltip-append-to-body="true" tooltip="Chart settings"><i class="icon-pie-chart"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'palette')" tooltip-append-to-body="true" tooltip="Palette"><i class="icon-palette"></i></div>
```
juliusv: I think this won't need a palette tab at all, right? Right now it just says that it's not implemented yet.

app/assets/templates/gauge_template.html.erb
```diff
@@ -0,0 +1,107 @@
...
+          <label>Refresh interval (seconds):</label>
+          <input class="form-control" type="number" ng-model="graph.refresh" min="1">
+          <label>Precision:</label>
+          <input class="form-control" type="number" ng-model="graph.precision" min="0">
+          <label>Warning threshold (percent):</label>
+          <input class="form-control" type="number" ng-model="graph.warning" max="100" min="1">
+          <label>Danger threshold (percent):</label>
+          <input class="form-control" type="number" ng-model="graph.danger" max="100" min="1">
+          <div class="checkbox">
+            <label><input type="checkbox" ng-model="graph.hideGauge" /> hide gauge</label>
```
juliusv: Maybe `Hide graphical gauge`?

app/assets/templates/gauge_template.html.erb
```diff
@@ -0,0 +1,107 @@
...
+  <div class="widget_wrapper panel panel-default pull-left">
+    <error-message error-messages="errorMessages"></error-message>
+    <div class="js_widget_wrapper widget_title panel-heading" ng-mouseenter="showControls = true" ng-mouseleave="showControls = false">
+      <span ng-show="dashboard"><a href="/{{dashboard.slug}}" class="dashboard_link pointer">{{dashboard.name}}</a> &mdash; </span>
+      {{title()}}
+
+      <div class="btn-group pull-right graph_config_menu">
+        <div ng-show="showControls || showTab || requestsInFlight" class="btn btn-primary btn-sm" ng-click="refreshGraph()" tooltip-append-to-body="true" tooltip="Refresh"><i ng-class="{spin:requestInFlight}" class="glyphicon glyphicon-refresh"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'datasources')" tooltip-append-to-body="true" tooltip="Datasources"><i class="icon-add-to-list"></i></div>
+        <div ng-show="showControls || showTab" class="btn btn-primary btn-sm" ng-click="toggleTab($event, 'graph')" tooltip-append-to-body="true" tooltip="Chart settings"><i class="icon-pie-chart"></i></div>
```
juliusv: s/icon-pie-chart/icon-gauge/

juliusv:
Looks pretty cool otherwise.

app/assets/templates/gauge_template.html.erb
stuartnelson3: yep, global refresh || 60s

app/assets/templates/gauge_template.html.erb
stuartnelson3: sure

app/assets/templates/gauge_template.html.erb
stuartnelson3: :+1:

stuartnelson3:
give the saving a try. the changes to endtime/date were propagating to the gauge, which was undesired behavior.

stuartnelson3:
@juliusv ping, how's it lookin

juliusv:
Sorry, so many reviews in the queue :-/

Just testing this again and noticed something I hadn't noticed earlier. The "Link to Prometheus" feature isn't working for gauges because the `linkToPrometheus()` function doesn't return a link. The exceptions made for the `pie` widget type in that function also need to be made for the `gauge` type to make it work. Also, the generated link needs to link to the tabular view, since scalar values cannot be graphed.

juliusv:
Two more things:

1)

When I do:

- Add Gauge
- start typing any letter in the expression input without selecting a server first

...I get JS errors:

```
TypeError: Cannot read property 'length' of undefined
    at application.js:1954
    at k.promise.then.w (application.js:1836)
    at k.promise.then.w (application.js:1836)
    at application.js:1837
    at h.$get.h.$eval (application.js:1845)
    at h.$get.h.$digest (application.js:1842)
    at h.$get.h.$apply (application.js:1845)
    at application.js:1855
    at e (application.js:1777)
    at application.js:1780
```

2) the usual "press enter in the expression input to reload the data" doesn't work for the gauge.

stuartnelson3:
@juliusv mk give 'er a try

juliusv:
Nice. One last thing: I think we discussed defaulting the refresh interval to the global dashboard default if it's omitted. But omitting it causes it to reload every 1s?

stuartnelson3:
ah, forgot about that. it's here:

https://github.com/prometheus/promdash/pull/365/files#diff-f0f83c1924b976d6c3a07f1e53d509b6R65

what should happen if neither the widget, or global refresh, is set? 1 second?

juliusv:
I'd say it should not refresh at all in that case...

stuartnelson3:
mk

stuartnelson3:
alright that ought to do it

juliusv:
Aaaaalmost there. Now I'm getting schema validation errors upon save. The relevant bits:

```
    - The property '#/widgets/1/refresh' of type NilClass did not match the following type: integer
    - The property '#/widgets/1' contains additional properties ["hideGauge"] outside of the schema when none are allowed
```

stuartnelson3:
aaaaaaaaaand plzzzzzzzz work

juliusv:
Hmm,

- refresh on enter in the expression input has stopped working again?
- now I can save, but the gauge does not load immediately when the dashboard loads, but only upon the next refresh cycle (this is with `30s` global refresh and empty refresh in the gauge widget)?

juliusv:
It appears that "refresh upon enter" only works when there is a refresh interval configured in the gauge.

stuartnelson3:
mk. got that.

if i missed something again, i'm deleting this pr.

grobie:
`s/if i missed something again, i'm deleting this pr./if i missed something again, I didn't write enough tests./`

juliusv:
So at least everything seems to work for me now :) So :+1:

PR #366, opened 2015-03-27
Fix double-escaping issue for multiple expressions.

juliusv:
When there were multiple expressions configured for a graph, each legend
string would be escaped once for every expression, leading to
double-escaping or worse.


juliusv:
@stuartnelson3

stuartnelson3:
:+1:

PR #367, opened 2015-03-31
Generate home URL using Ruby

quinox:
This will generate the homepage URL instead of hardcoding it.

I hacked my PromDash to use a specfic path prefix so I can use it behind a proxy. A hardcoded link to `/` breaks this setup.


juliusv:
:+1: Thanks!

Issue #309, opened 2015-01-29
Show metric color in value tooltip


grobie:
When hovering over values in a graph with many metrics or metrics, it's sometimes hard to see for which metric the tooltip is shown. While the circle changes its color already, it's sometimes hidden behind the mouse pointer for example.

Suggestion: show the color of the metric again inside of the tooltip. Possibilities: a) use the color as background b) add a small square with the color and put it before/after the value time/name c) something else.

grobie:
Actually, it would be cool to show the legend for that time series in the tooltip as well. That would make some graphs a lot clearer.

juliusv:
:+1: especially because the legend sometimes only contains information which is not actually from Prometheus labels, but which indiciates which expression a time series is from. If you e.g. add two of the same expressions, but from different Prometheus servers, you might get two identical sets of time series that you can't differentiate via the hover detail currently (but via different legend format strings).

grobie:
So what about option b) plus a title. Maybe `▢ <server name>: <label string>`?

PR #368, opened 2015-04-01
[WIP] Show color and series name in hover detail

grobie:
Solves #309.


grobie:
@juliusv I thought this can't be that hard. It turns out it was super easy, you seem to have done half of the work years ago already. Lingering code FTW eh ;)

I want to iterate a bit more on the box, but having the color and series name there is already super great!

promdash - hoverdetail

juliusv:
I like this in general, but it needs some tuning:

When there is no custom legend string, the full time series name is shown at the top, which is way too long. We could only show the metric name there in that case, but that would be somewhat redundant because that name is already shown right below the date (when there is a metric name at all). And what to show when there is no metric name *and* no legend string? All these different cases need to look sane somehow.

* when there is a legend string: show legend string
* no legend string, but metric name: show metric name (even if it's redundant? Or consolidate somehow?)
* no legend string, no metric name: maybe put the colored box infront of the `value: ...`?

PR #369, opened 2015-04-01
Configurable path prefix to run PromDash behind a proxy

quinox:
Same as https://github.com/prometheus/prometheus/pull/614 but then for PromDash.

This pull request should be closely reviewed, I have no experience with Rails. All feedback greatly appreciated.

According to `rake routes` everything is covered. I am currently running PromDash on this code behind a proxy without any issues.

* I had to delete the `tmp/` folder to flush some cached files. I have no idea if this is always required.
* `graph_template.html.erb` and `pie_template.html.erb` contain an `<a href="/{{dashboard.slug}}">` but I have no idea how those tags are used

### DIFF
--- a/app/assets/javascripts/angular/controllers/dashboard_ctrl.js.erb
+++ b/app/assets/javascripts/angular/controllers/dashboard_ctrl.js.erb
@@ -32,7 +32,7 @@ angular.module("Prometheus.controllers")
       return;
     }
     $scope.generatingPermalink = true;
-    $http.post($window.location.origin + "/permalink", buildDashboardJSON()).then(function(payload) {
+    $http.post($window.location.origin + "<%= Rails.configuration.path_prefix %>permalink", buildDashboardJSON()).then(function(payload) {
       $scope.showPermalink = true;
       $scope.permalink = $window.location.origin + payload.data.url;
       var input = $("[ng-model=permalink]")[0];
@@ -189,7 +189,7 @@ angular.module("Prometheus.controllers")
 
   $scope.queryDirectory = function() {
     $scope.directoryForClone.id = $scope.directoryForClone.id || "unassigned";
-    $http.get('/directories/' + $scope.directoryForClone.id).then(function(payload) {
+    $http.get('<%= Rails.configuration.path_prefix %>directories/' + $scope.directoryForClone.id).then(function(payload) {
       $scope.dashboardNames = payload.data.dashboards;
       $scope.dashboardForClone = payload.data.dashboards.filter(function(d) {
         return d.name == dashboardName;
@@ -199,7 +199,7 @@ angular.module("Prometheus.controllers")
   };
 
   $scope.queryDashboard = function() {
-    $http.get('/dashboards/' + $scope.dashboardForClone.id + '/widgets').then(function(payload) {
+    $http.get('<%= Rails.configuration.path_prefix %>dashboards/' + $scope.dashboardForClone.id + '/widgets').then(function(payload) {
       $scope.dashboardWidgets = payload.data.widgets;
       $scope.widgetToClone = payload.data.widgets[0];
     });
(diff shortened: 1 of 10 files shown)


juliusv:
@quinox Sorry, I'll probably only get to this in >1 week due to vacation. Or maybe @stuartnelson3 wants to take a look (if yes, please verify that all instances of using the prefix actually work). In any case, PromDash is a slow-moving target at the moment, so I would expect everything to still apply without conflict in a week.

quinox:
No problem, my own system is working fine so there's no rush. The patch is rather straight forward, so easy to fix if it goes into conflict.

juliusv:
Getting to this now. To be sure that things still apply cleanly, could you rebase ontop of current master?

quinox:
Squashed and rebased.

juliusv:
Thanks, taking a look now.

juliusv:
Great, most of this seems to be working fine, but I still found three areas where it's using the unprefixed path and thus gets a `404` / breaking functionality:

- "Link to Graph" button on any widget (`/w`)
- "Clone Widget" (`/directories.json`)
- "Create permalink to current dashboard state" (`/permalink`)

app/assets/javascripts/angular/controllers/dashboard_ctrl.js.erb
```diff
@@ -189,7 +189,7 @@ angular.module("Prometheus.controllers")
 
   $scope.queryDirectory = function() {
     $scope.directoryForClone.id = $scope.directoryForClone.id || "unassigned";
-    $http.get('/directories/' + $scope.directoryForClone.id).then(function(payload) {
+    $http.get('<%= Rails.configuration.path_prefix %>directories/' + $scope.directoryForClone.id).then(function(payload) {
```
juliusv: The prefix usages in this file all need a slash after the path prefix to work.

app/assets/javascripts/angular/services/annotation_refresher.js.erb
```diff
@@ -13,7 +13,7 @@ angular.module("Prometheus.services").factory('AnnotationRefresher', ["$http", "
       var until = Math.floor(graph.endTime || Date.now()) / 1000;
 
       tags.forEach(function(t) {
-        $http.get('/annotations', {
+        $http.get('<%= Rails.configuration.path_prefix %>annotations', {
```
juliusv: Same here (slash separator missing).

app/assets/javascripts/angular/services/widget_link_helper.js.erb
```diff
@@ -1,7 +1,7 @@
 angular.module("Prometheus.services").factory('WidgetLinkHelper', ["$http", "InputHighlighter", function($http, InputHighlighter) {
   return {
     createLink: function(data) {
-      this.promise = $http.post('/w', data);
+      this.promise = $http.post('<%= Rails.configuration.path_prefix %>w', data);
```
juliusv: Dito here and below.

juliusv:
Hmm, actually, it does work when I end the prefix itself with a slash (thought at first it didn't, but must have done something wrong). I wonder, could we auto-sanitize the prefix somehow so that it always contains a slash, no matter what the config provides?

juliusv:
Ah yeah, the reason I first thought it didn't work at all even with a slash in the prefix were the cached temp files, I think.

juliusv:
But definitely, the dashboard permalinking (the little link button at the top in the global settings) needs a prefix added in these two places:

- https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/controllers/dashboard_ctrl.js#L35
- https://github.com/prometheus/promdash/blob/master/app/controllers/dashboards_controller.rb#L51

quinox:
I sanitized path_prefix inside `environment.rb`, seems like a good spot (but again, I don't know any Rails).

quinox:
Note to self: Remove settings from `environments/*.rb`, set them inside `config/application.rb` based on ENV and auto-update `config.path_prefix` (when needed)

quinox:
Implemented the move from config file to ENV variable.

In my research I came across an existing ENV variable named `RAILS_RELATIVE_URL_ROOT`. The documentation about this variable talks about deploying to another directory which this feature isn't about (`path_prefix` is purely for routing). When I used it it also messed up serving static files so I decided to use a different variable name.

juliusv:
Excellent, everything I tried worked! :+1: Could you squash it? Then I'll merge.

quinox:
It's all prepped.

juliusv:
Thanks!

Issue #180, opened 2014-07-16
Support multiple servers for the same dashboard


grobie:
As a developer,
Given I deploy my services in multiple zones,
I want to have one dashboard
And switch between the data collected by different prometheus servers in each zone.

mwitkow-io:
A great way to solve it would be through adding template variables in server URLs. It's becoming a real maintenance burden not having this functionality.

juliusv:
@mwitkow-io Agreed on being a maintenance burden (for us too!), and on doing it via template variables. Now we just have to figure out a good UI and JSON encoding for doing that. Currently servers for each expression are saved by their ID. Now they'd need to (optionally) be referred to by a string with some template variables inside instead, which then resolves (hopefully) to an existing server. Biggest challenge is the UI, I think: how to keep the simple case simple with just a dropdown, as it is now, but make it easy to use template variables for servers in individual graphs/expressions instead? I guess which type of UI to use would not be a per-graph/per-expression setting, but a global one (as otherwise it'd become quite cumbersome). So maybe it could be just a global setting which chooses the type of UI to use? Hrm... better ideas?

matthiasr:
Maybe add the templating only on the _URL_ side of the server, then automatically import all variables there into the dashboard and from that point on treat them as if they were a dashboard variable?

matthiasr:
i.e. you'd have one server definition for "prometheus-routing", pointing at `http://prometheus-routine-{{zone}}:9090`, then once you use that in a dashboard that dashboard automatically grows a `zone` template variable.

Obviously you could not use different `zone` variables for other things in the dashboard then but you could use the same one. I think that is a good thing.

juliusv:
@matthiasr Oh, that's a great idea! Thank you. I'm sold! :)

And it even keeps PromDash completely agnostic about the concept of a `zone`. It'd just be an arbitrary template variable used in the server name.

mwitkow-io:
Is anyone working on this? @matthiasr

stuartnelson3:
this issue is not currently being worked on

mwitkow-io:
K, I think we'll hack on it in our spare time.

Can you drop some pointers for "hacking noobs" as to where to start? :)

stuartnelson3:
[...]

PR #372, opened 2015-04-12
Vars Interpolation in Servers

mwitkow-io:
So this is a first pass on https://github.com/prometheus/promdash/issues/180. I'm quite foreign to AngularJS and Ruby, so I'm not sure if this is the right approach.

Everything seems to work:
 * specifying `{{my_var}}` in a server URL will make it be interpolated from template `scope.vars`
 * if template vars already don't specify the variable, it will be extracted out of server definitions and added with an empty value

There is one problem now:
Listing and viewing servers that contain interpolation variables causes them to be ignored in HTML. Editing in HTML and viewing in JSON works, and the data is in the DB. I presume (wild guess here) the problem may be that AngularJS interprets `{{my_var}}` as a directive and tries to bind it, and it defaults to empty. I don't know how to work around it though. Is there a way to make AngularJS ignore directives in a given node?

### DIFF
--- a/app/assets/javascripts/angular/resources/shared_graph_behavior.js
+++ b/app/assets/javascripts/angular/resources/shared_graph_behavior.js
@@ -60,6 +60,22 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
       }
     }
 
+    function addVarsFromServerInterpolations(servers) {
+      var singleInterpolation = /{{\s?(\w+)\s?}}/g;
+      for (var i = 0; i < servers.length; i++) {
+        while (match = singleInterpolation.exec(servers[i].url)) {
+          var_name = match[1]
+          if (!(var_name in $scope.globalConfig.vars)) {
+            console.log("Found " + var_name + " in " + servers[i].url)
+            $scope.globalConfig.vars[var_name] = ''  // no default value
+          }
+        }
+      }
+    }
+
+    $scope.servers = servers;
+    addVarsFromServerInterpolations(servers)
+
     $scope.$watch(function() {
       return $location.url();
     }, function() {
@@ -67,7 +83,7 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
     });
 
     $scope.globalConfig.tags = $scope.globalConfig.tags || [];
-    $scope.servers = servers;
+
 
     $scope.sortableOptions = {
       handle: ".widget_title",
--- a/spec/javascripts/angular/services/url_generator_spec.js
+++ b/spec/javascripts/angular/services/url_generator_spec.js
@@ -8,7 +8,7 @@ describe('URLGenerator', function() {
   it('allows urls with custom paths, no trailing slash', function() {
     ['http://promdash.server.com/prometheus', 'http://promdash.com'].forEach(function(s) {
       ['/api/query_range', '/api/query', '/api/metrics', '/arbitrary/endpoint'].forEach(function(ep) {
-        var url = urlGenerator(s, ep);
+        var url = urlGenerator(s, ep, {});
         expect(url).toEqual(s + ep);
       });
     });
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -60,6 +60,22 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
...
+          var_name = match[1]
+          if (!(var_name in $scope.globalConfig.vars)) {
+            console.log("Found " + var_name + " in " + servers[i].url)
+            $scope.globalConfig.vars[var_name] = ''  // no default value
+          }
+        }
+      }
+    }
+
+    $scope.servers = servers;
```
mwitkow-io: @stuartnelson3, I have no idea where `servers` comes from. It's neither from `$scope` nor directly injected. Is this some magical global variable?

stuartnelson3: yeah, unfortunately it's being rendered into the dashboard#show template and made globally available.

https://github.com/prometheus/promdash/blob/master/app/views/dashboards/show.html.erb#L5

stuartnelson3: it would be cool to not have the unnecessary `while`, i think it's potentially misleading because it doesn't actually loop in this case. `var_name` was also an unintentional global, and since it is so short lived I prefer single character variable names.

```javascript
function addVarsFromServerInterpolations(servers) {
  var re = /{{\s?(\w+)\s?}}/g;
  for (var i = 0; i < servers.length; i++) {
    var m = re.exec(servers[i].url);
    if (m && !$scope.globalConfig.vars[m[1]]) {
      $scope.globalConfig.vars[m[1]] = '';
    }
  }
}
```

stuartnelson3:
Thanks for the pr!

couple things:

1. Your build failed because of jshint. You can check travis to see what failed or run it locally (instructions in readme)
2. the `server#index` and `server#show` pages aren't rendering correctly because the `{{}}` interpolation in server names gets stripped out by angular (well, angular interpolates the value it has for whatever is in the `{{}}`, which is nothing). you can fix this by:

`servers/index.html.erb`
```html
  <tbody ng-non-bindable>
    <!-- unmodified stuff -->
  </tbody>
```
`servers/show.html.erb`
```html
<p ng-non-bindable>
  <strong>URL:</strong>
  <%= @server.url %>
</p>
```
(http://stackoverflow.com/questions/15674508/prevent-angularjs-interpolation-in-certain-dom-hierarchies)

@juliusv thoughts? page load currently errors if there is a widget depending on this but no value is interpolated, z.B. `our-awesome-server-{{zone}}.com` `->` `our-awesome-server-.com`, but that could happen to any dashboard depending on a particular url var.

spec/javascripts/angular/services/url_generator_spec.js
```diff
@@ -17,7 +17,7 @@ describe('URLGenerator', function() {
   it('allows urls with custom paths, with trailing slash', function() {
     ['http://promdash.server.com/prometheus/', 'http://promdash.com/'].forEach(function(s) {
       ['/api/query_range', '/api/query', '/api/metrics', '/arbitrary/endpoint'].forEach(function(ep) {
-        var url = urlGenerator(s, ep);
+        var url = urlGenerator(s, ep, {});
         expect(url).toEqual(s.substring(0, s.length - 1) + ep);
       });
     });
```
stuartnelson3: could you add a test checking that the interpolation is going into the url correctly?

```javascript
  it('interpolates into urls', function() {
    var s = 'http://promdash.{{zone}}.com/prometheus';
    ['aa', 'bb', 'cc', 'dd'].forEach(function(zone) {
      var url = urlGenerator(s, '', {zone: zone});
      expect(url).toEqual('http://promdash.' + zone + '.com/prometheus');
    });
  });
```

juliusv:
Can this one be closed in favor of https://github.com/prometheus/promdash/pull/373?

grobie:
Great! This solves a great deal of #180 already!

PR #373, opened 2015-04-13
Server interpolation

stuartnelson3:
@mwitkow-io since i had already gone through with the work i just pushed up the branch but maintained your commits, hope you don't mind

@juliusv

### DIFF
--- a/app/assets/javascripts/angular/resources/shared_graph_behavior.js
+++ b/app/assets/javascripts/angular/resources/shared_graph_behavior.js
@@ -60,6 +60,22 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
       }
     }
 
+    function addVarsFromServerInterpolations(servers) {
+      var singleInterpolation = /{{\s?(\w+)\s?}}/g;
+      for (var i = 0; i < servers.length; i++) {
+        while (match = singleInterpolation.exec(servers[i].url)) {
+          var_name = match[1]
+          if (!(var_name in $scope.globalConfig.vars)) {
+            console.log("Found " + var_name + " in " + servers[i].url)
+            $scope.globalConfig.vars[var_name] = ''  // no default value
+          }
+        }
+      }
+    }
+
+    $scope.servers = servers;
+    addVarsFromServerInterpolations(servers)
+
     $scope.$watch(function() {
       return $location.url();
     }, function() {
@@ -67,7 +83,7 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
     });
 
     $scope.globalConfig.tags = $scope.globalConfig.tags || [];
-    $scope.servers = servers;
+
 
     $scope.sortableOptions = {
       handle: ".widget_title",
--- a/app/assets/javascripts/angular/resources/shared_graph_behavior.js
+++ b/app/assets/javascripts/angular/resources/shared_graph_behavior.js
@@ -61,20 +61,22 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
     }
 
     function addVarsFromServerInterpolations(servers) {
-      var singleInterpolation = /{{\s?(\w+)\s?}}/g;
+      var re = /{{\s?(\w+)\s?}}/g;
+      var m;
       for (var i = 0; i < servers.length; i++) {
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -60,6 +60,19 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
       }
     }
 
+    function addVarsFromServerInterpolations(servers) {
+      var re = /{{\s?(\w+)\s?}}/g;
+      for (var i = 0; i < servers.length; i++) {
+        var m = re.exec(servers[i].url);
```
mwitkow-io: I see you removed the `while`. The `re.exec` is a weird API, returning next match until it returns `null` when it can't find any more.
Without the `while`, only the first match will be found and only one interpolation will be found.

mwitkow-io:
Thanks for fixing my deficiencies :)
Just one thing regarding `addVarsFromServerInterpolations`.

app/assets/javascripts/angular/resources/shared_graph_behavior.js
stuartnelson3: ooooh, interesting. thanks for following up on that and correcting me!

mwitkow-io:
LGTM :) Thanks!

juliusv:
Awesome. There are still some places where interpolations in the server name or URL get interpreted as Angular expressions:

- create a new server with name `foo-{{zone}}` and URL `http://foo-{{zone}}`
 - the resulting "success" message will lack the `{{zone}}` part
 - when editing the server again, the input fields will lack the `{{zone}}` part

Hmm, I think there might be other places where this happens, independently of this PR. Same for dashboard names, for example. Angular escaping for this kind of scenario has never been on our radar so far... basically, whenever a string is put from the database into the view server-side (vs. inserted via JS after page load), any Angularisms in it would have to be escaped :-/

stuartnelson3:
we could either

1. play whack-a-mole with `{{}}` in our views
2. only load angular on `dashboard#show.html.erb`

preference?

juliusv:
Yeah, restricting Angular to only the necessary single view probably makes sense.

I also found this:
https://github.com/angular/angular.js/issues/5601

It looks like escaping any `{{...}}` to `\{\{\...}\}` on the server side would do the proper thing. (see also https://docs.angularjs.org/api/ng/service/$interpolate). That's an alternative to using `ng-non-bindable`. Not sure if one is much better than the other. The recommendation on that issue thread actually is to not mix server-side and client-side template engines for this reason (https://github.com/angular/angular.js/issues/5601#issuecomment-31548902).

For now, either approach seems fine to me.

stuartnelson3:
@juliusv split the javascript into two separate files that are conditionally loaded, and added some specs (apparently i never added server specs previously)

mwitkow-io:
Is there anything left to be fixed here? :) We'd really appreciate if this was merged in :)

juliusv:
Well, at the moment, it seems broken to me - the title and other things in graphs are not interpreted correctly by AngularJS anymore:

promdash_broken

@stuartnelson3 Did you add all necessary files to git?

stuartnelson3:
oh I added all the necessary files, but I put `ng-non-bindable` on the graph title instead of the dashboard title :P

pushed, should be copasetic

juliusv:
Great, that works now. Only thing left that I could find is that the interpolations still happen in the non-show views of the dashboards controller (e.g. when renaming a dashboard with `{{}}` in the name). That still needs to be fixed.

app/helpers/application_helper.rb
```diff
@@ -9,6 +9,11 @@ def page_title(title = nil)
     end
   end
 
+  def widget_page?
+    ["single_widget", "embed"].include?(controller_name) ||
+      "dashboards_show" == "#{controller_name}_#{action_name}"
```
stuartnelson3: @juliusv made it specifically check for dashboards_show, was easier at this point

juliusv:
Works! Woot! :+1:

stuartnelson3:
thanks for your patience! will merge once tests pass.

grobie:
Great! This solves a great deal of #180 already!

PR #374, opened 2015-04-13
Set pie charts to query and aggregate data from multiple servers.

stuartnelson3:
turn it up to 11

@juliusv @grobie 

**TODO**

- [x] json schema
- [x] fix pie fixture json

### DIFF
--- a/app/assets/javascripts/angular/directives/pie_chart.js
+++ b/app/assets/javascripts/angular/directives/pie_chart.js
@@ -18,7 +18,7 @@ angular.module("Prometheus.directives").directive('pieChart', ["$location", "Wid
         $el.css('height', graphHeight);
 
         if (pieGraph) {
-          $el.html('<div class="graph_chart"></div>');
+          $el.html('<div class="graph_chart pie_chart"></div>');
           pieGraph = null;
         }
 
@@ -26,20 +26,37 @@ angular.module("Prometheus.directives").directive('pieChart', ["$location", "Wid
           return;
         }
 
-        pieData.forEach(function(e) {
-          e.value = parseFloat(e.value);
+        var expressionIDToLegendID = {};
+        scope.graphSettings.expressions.forEach(function(expr) {
+          expressionIDToLegendID[expr.id] = expr.legendID;
+        });
 
-          if (scope.graphSettings.legendFormatString) {
-            e.ts = VariableInterpolator(scope.graphSettings.legendFormatString, e.metric);
+        legendIDToString = {};
+        scope.graphSettings.legendFormatStrings.forEach(function(ls) {
+          legendIDToString[ls.id] = ls.name;
+        });
+
+        pieData.forEach(function(e) {
+          var m = e.metric;
+          m.value = e.value = parseFloat(e.value);
+
+          var legendID = expressionIDToLegendID[e.expressionID];
+          var legendStr;
+          if (legendID) {
+            legendStr = legendIDToString[legendID];
+          }
+
+          if (legendStr) {
+            e.ts = VariableInterpolator(legendStr, m);
           } else {
-            var ts = joinProperties(e.metric, "=").map(function(t) {
+            var ts = joinProperties(m, "=").map(function(t) {
               return t.replace(/=(.+)/, function($1, $2) {
                 return "=\"" + $2 + "\"";
               });
             }).join(",");
-            e.ts = (e.metric.__name__ || '') + "{" + ts + "}";
+            e.ts = (m.__name__ || '') + "{" + ts + "}";
           }
-          tooltip[e.ts] = e.metric;
+          tooltip[e.ts] = m;
         });
 
         var svg = dimple.newSvg($el.find(".graph_chart")[0], graphWidth, graphHeight);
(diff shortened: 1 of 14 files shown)


grobie:
Again the question why a pie chart needs to be its own widget type. It's a bit unexpected from my user perspective to need to create a whole new widget just to visualize the same data I already have. Is it possible to merge it back?

stuartnelson3:
with this PR, they are now using the same graph refresher service object, and merging them should™ be straightforward. I can look at it tomorrow.

to clarify: you're interested in merging the two together because you want to be able to go into your graph settings tab, click "pie" as the visualization type, and get fast feedback on the most recent data?

stuartnelson3:
@juliusv thoughts? ^^

grobie:
Correct, while creating a new dashboard/graph I want to see how the data
looks like using the different visualization types before opting for one.
Being able to quickly switch between them makes that a lot easier.

juliusv:
@grobie You mean graph, pie chart (and the new gauge?) should all be one widget type? I dunno, maybe. Half of the configuration for them is different (mainly the graph has many more options which don't apply to the pie chart). I guess it could be done by converting the underlying widget JS object from one widget type to another. Switching from graph to pie chart would then forget all the graph options though, so you couldn't easily switch back and forth while preserving settings that only apply to one of the types. Maybe it's worth it, maybe not - up to you guys I'd say :)

stuartnelson3:
@grobie @juliusv I'm going to nix the pie/graph merging for now. Managing the JSON schema as a big mashup between the two sounds unappealing, but as j mentioned, to convert from one widget type to another would lose all configuration data, completely ruining what you're asking for.

Let's move the merge to a separate PR. Since I have other responsibilities, I don't want these branch to languish.

grobie:
Fair enough.

stuartnelson3:
@juliusv @grobie how's this one look

juliusv:
Soorrry for the long PR queue lag :-/

Was just giving this a spin now and discovered some initial problems:

- the `x` delete button on an expression doesn't do anything
- when there are multiple expressions, the legend format strings also need to be multiple, and assignable (exactly as in graphs)

stuartnelson3:
k made those changes

let me know what's up, went through it and it *seems* to be working

juliusv:
Hmm, the format strings don't work for me correctly:

1. create two expressions from the same server, e.g. `count(up)` and `count(prometheus_local_storage_chunk_ops_total)`.
2. create two legend format strings, `a` and `b`.
3. assign `a` to the first expression and `b` to the second.
4. the pie chart only contains the second expression and the legend says only `b`

juliusv:
Also, the `x` button now closes the menu tab pane as a side-effect.

stuartnelson3:
>Also, the x button now closes the menu tab pane as a side-effect.

that's a pre-existing bug (check in our current deploy)

stuartnelson3:
I can confirm that the legend format strings aren't working correctly. I had been testing with a different server in each expression.

app/assets/javascripts/angular/resources/shared_graph_behavior.js
```diff
@@ -241,7 +241,7 @@ angular.module("Prometheus.services").factory("SharedGraphBehavior", ["$http", "
     $document.on('click', function(ev) {
       // If the following selectors aren't in the event.target's ancestor chain,
       // clear open menus from the screen.
-      if (!$(ev.target).closest('.graph_config_menu, .graph_control_tabbar, #global_controls, .add_widget_button, .modal_container').length) {
+      if (!$(ev.target).closest('.graph_config_menu, .graph_control_tabbar, #global_controls, .add_widget_button, .modal_container, .expression_select').length) {
```
stuartnelson3: fixes the errant closing when trying to remove an expression

app/assets/javascripts/angular/services/graph_refresher.js
```diff
@@ -31,6 +26,14 @@ angular.module("Prometheus.services").factory('GraphRefresher',
               'data': data
             });
             break;
+          case 'vector':
+            var d = data.value || data.Value;
+            d.forEach(function(s) {
+              s.metric.serverName = server.name;
+              s.expressionID = expressionID;
```
stuartnelson3: scopes the legend string to an expression. previously it was to the server, which you discovered caused a bug.

stuartnelson3:
@juliusv updated

app/assets/javascripts/angular/directives/pie_chart.js
```diff
@@ -26,20 +26,37 @@ angular.module("Prometheus.directives").directive('pieChart', ["$location", "Wid
...
+        });
+
+        legendIDToString = {};
+        scope.graphSettings.legendFormatStrings.forEach(function(ls) {
+          legendIDToString[ls.id] = ls.name;
+        });
+
         pieData.forEach(function(e) {
-          e.value = parseFloat(e.Value || e.value);
+          var m = e.Metric || e.metric;
```
grobie: How long do we need to support these different spellings for?

stuartnelson3: I believe we're past the 1 week mark:

https://github.com/prometheus/promdash/commit/6c7d370eae0aa82792fddbf9a687d6a73b1e194f

juliusv: Haha, yes. I volunteer to clean these up unless you already started.

Issue #381, opened 2015-04-17
Support grafana color schemes


grobie:
Grafana has actually some nice color schemes. Let's add them! http://play.grafana.org/

PR #384, opened 2015-04-17
shamelessly steal grafana color schemes

stuartnelson3:
:tada: 

requested in #381 

note: do we still want the old color schemes? i kept spectrum2001 cause it's cool, but no one really used any others.

@grobie @juliusv

### DIFF
--- a/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/dashboard_ctrl.js
@@ -62,7 +62,7 @@ angular.module("Prometheus.controllers")
   $scope.widgets = $scope.widgets || dashboardData.widgets || [];
   SharedGraphBehavior($scope, dashboardData.globalConfig);
   $scope.palettes = Palettes;
-  $scope.globalConfig.palette = $scope.globalConfig.palette || 'colorwheel';
+  $scope.globalConfig.palette = $scope.globalConfig.palette || 'spectrum2001';
   $scope.globalConfig.resolution = $scope.globalConfig.resolution || 4;
 
   $scope.saveDashboard = function() {
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -51,6 +51,7 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       id: id,
       serverID: serverID,
       axisID: axisID,
+      palette: '',
       expression: ''
     };
     $scope.graph.expressions.push(exp);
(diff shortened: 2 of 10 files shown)


stuartnelson3:
also, you can now set color schemes per expression.

juliusv:
:+1: for the branch name. Now looking at this closer :P

Do you know what's up with the failing tests?

stuartnelson3:
intermittent failure, happens all the time

juliusv:
Ok, so three things I see until now:

- what do "m" and "g" stand for? It seems the "g" ones contain variants of "all" the colors, while the "m" ones contain variants of the same one? Some better names would be good. At least for the latter, one could put the name of the base color into the name? Otherwise the user just sees a long list of "<one-letter><number>" combinations, which doesn't tell them anything.
- I'm wary of simply throwing out the old color schemes... especially right now, they just get replaced with a random new one, which can throw people off. Unfortunately, we don't know what people have been using out there. Maybe just keep them all? :-/
- The feature to set the color scheme per expression is interesting, but the expression input was already way too small before, and now it's completely unusably small due to the new input adjacent to it. Mabye hold off on it until we have a better solution? AFAIR the primary thing that people requested was actually once to set the color, not the colorscheme, for individual expressions, right?

stuartnelson3:
>what do "m" and "g" stand for? It seems the "g" ones contain variants of "all" the colors, while the "m" ones contain variants of the same one? Some better names would be good. At least for the latter, one could put the name of the base color into the name? Otherwise the user just sees a long list of "" combinations, which doesn't tell them anything.

"m" and "g" don't stand for anything, they're just random letters. the "g" color schemes were stolen from grafana, and "m" is created programmatically by selecting all the similar colors from "g" (check out `palettes.js`)

>I'm wary of simply throwing out the old color schemes... especially right now, they just get replaced with a random new one, which can throw people off. Unfortunately, we don't know what people have been using out there. Maybe just keep them all? :-/

yeah I'm fine keeping them all, no big deal

>The feature to set the color scheme per expression is interesting, but the expression input was already way too small before, and now it's completely unusably small due to the new input adjacent to it. Mabye hold off on it until we have a better solution? AFAIR the primary thing that people requested was actually once to set the color, not the colorscheme, for individual expressions, right?

is it a UI issue, or a feature issue? While people wanted to set individual colors (e.g. status codes), they could now select one of the "m" colors that only has one type and set it for an expression that only e.g. returns 500s.

juliusv:
- Ok, I can take a look at helping with naming these properly later
- Ok, then the scheme-per-expression is mostly a UI issue. Besides squishing the expression input, it feels like the per-expression inputs are getting too complex in general. Too many different selection boxes. I just don't know if this feature is worth the extra input at the moment. Open for ideas of accomodating both concerns better, of course...

grobie:
Let's keep the old ones! They're also used for sure, here a summary from SoundCloud (ignoring all widgets using the default profile without an explicit selection):

```
classic9: 19
colorwheel: 130
cool: 25
munin: 121
spectrum14: 20
spectrum2000: 1
spectrum2001: 45
```

I'd need to see the different colors first, but maybe "Grafana $type" e.g. "Grafana dark" and `Grafana bright" or so?

grobie:
Let's just switch to Grafana :)

PR #390, opened 2015-04-30
Don't add frame if user cancels.

### LINKED ISSUE #388: Cancelling the insertion of a frame breaks the UI
Tested against HEAD (98f955c):

* Open a dashboard
* Click on "Add frame"
* Press cancel
* Get a stacktrace in your terminal starting with: `Error: url is null`
* The newly created elements can't be interacted with.

broken promdash

stuartnelson3:
Previously, canceling would not abort adding a frame, which would break
the UI.

fixes #388

### DIFF
--- a/spec/features/graph_spec.rb
+++ b/spec/features/graph_spec.rb
@@ -60,16 +60,23 @@ feature "Dashboard#show", js: true do
           open_tab 'Remove graph'
           click_button 'Delete'
           click_button 'Add Frame'
-          accept_alert
-          open_tab 'Remove frame'
+        end
+
+        it "should allow canceling on frame creation" do
+          dismiss_alert
+          expect(all('.widget_title').count).to eq 0
         end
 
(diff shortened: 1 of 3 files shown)


grobie:
:+1: (while a test would have been nice :p)

spec/features/graph_spec.rb
```diff
@@ -60,16 +60,23 @@
           open_tab 'Remove graph'
           click_button 'Delete'
           click_button 'Add Frame'
-          accept_alert
-          open_tab 'Remove frame'
+        end
+
+        it "should allow canceling on frame creation" do
+          dismiss_alert
```
stuartnelson3: @grobie
image

juliusv:
Woot :+1:

spec/features/graph_spec.rb
grobie: :dancers:

PR #398, opened 2015-05-07
Add container-fluid container class and apply to markup container.

stuartnelson3:
@grobie @juliusv

### DIFF
--- a/app/assets/stylesheets/dashboards.css.scss
+++ b/app/assets/stylesheets/dashboards.css.scss
@@ -16,6 +16,13 @@ $light-secondary-background: #f0f0f0;
   margin-bottom: 8px;
 }
 
+.container-fluid {
+  padding-right: 10px;
+  padding-left: 10px;
+  margin-right: auto;
+  margin-left: auto;
+}
+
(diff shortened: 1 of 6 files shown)


juliusv:
This makes the dashboard and server overview pages look broken (almost no margins at all).

stuartnelson3:
hm, so only on the `dashboard#show` page?

grobie:
Yes, this was the reason I didn't publish it.

juliusv:
Yeah, it should only be on the `dashboard#show` page. Also, do we have consensus on being ok with making non-fullscreen dashboards suddenly look completely different for people? I guess it's ok because it doesn't affect the wall-screen fullscreen view...

grobie:
Fine with me to have it only on dashboard#show as well instead of fixing the margin on the other pages.

juliusv:
SGTM too

stuartnelson3:
failed test is re-running, but all should be good.

even wrote a couple tests for the test ~~nazi~~ fairy

grobie:
:+1: great work!

PR #401, opened 2015-05-12
Add support for dashboard profiles

stuartnelson3:
@juliusv @grobie 

open questions:

- do we want limit the size of the dropdown profile selector?
- do we want to enforce that profile names have no spaces?

@grobie add any others you have


app/views/dashboards/show.html.erb
```diff
@@ -12,6 +13,18 @@
   </h1>
 
   <div id="global_controls">
+    <div ng-if="!globalConfig.showProfiles" class="global_grouping">
```
stuartnelson3: should this dropdown be hidden if no profiles exist?

grobie:
Changing profile names is a bit confusing right now. I guess we could even get away without this explicit Confirm/Cancel name change buttons, but otherwise `Enter` should at least confirm as well and leave the focus.

grobie:
Disabling legend items is not persisted across profiles. Given I have 3 profiles "db, dd, df", when I disable one metric by clicking on the legend string in profile db and switch to another profile, then the setting is not persisted. How hard would it be to use the same setting across all profiles here?

grobie:
Do we need to keep the "Show template variables" part? Might be confusing if profiles are used, as profiles will always win. Though, if someone doesn't want to use profiles, we should probably still provide the option for simple variables. Humm, how difficult would it be to display an info/warning box above the template variables in case any profiles exist? Something like "Global template variables are not available when profiles are used". And as i-Tüpfelchen disable all the form elements in this case as well.

grobie:
Yeah, unsure about spaces in profile names. It's cool to have them, but then we'd need to think about slugs and how to create+update them automatically. Maybe limiting profile names to `^[a-z0-9-_.]+$` is the easiest for now.

stuartnelson3:
I don't think we can persist legend items across profiles. Switching profiles could cause entirely different graphs to be rendered, ones that don't have a 1:1 relationship between time series. example would be if someone comes up with a profile that actually changes the expression completely (instead of switching between, e.g. two difference datacenters).

clearly this is not the intended use of profiles. should we ignore this, and assume that users will be using profiles for analogous graphs?

stuartnelson3:
>Yeah, unsure about spaces in profile names.

hm, what do you mean by create and update slugs? profile names are read as part of the route in the controller, and then when the page is loaded, the javascript code attempts to match the profile name from the route against any of the saved profiles and use those variables. if no match is found, the first saved profile is used. i think it would be much more work to "fix" users' profile name inputs to match our regex?

stuartnelson3:
>Do we need to keep the "Show template variables" part? Might be confusing if profiles are used, as profiles will always win.

I agree. I think we should show a warning, as you mention. It shouldn't be too hard, just a simple check to see if `profiles.length > 0`, and if so show some html.

I don't know why, but someone might want to still use template variables that are separate from profiles. there will also be a short migration period where people are transitioning to using profiles, and we don't want to wipe out their template variables.

I'm wary of disabling the template variables because lazy people might continue on using them and just add profiles to their dashboards.

how about I add the HTML warning, and we can talk more about disabling/removing the template variables?

stuartnelson3:
>Changing profile names is a bit confusing right now.

I remember now why the explicit edit/confirm/cancel buttons exist: 

Profiles are stored as key-value pairs. 
```javascript
{ 'profile-name': [ 
    { 'profile-variable-name': 'value'},
    {},
    {},
  ],
  'profile-name-2': [
    // ...
  ],
}
```

when changing `profile-name`, there was weird behavior because the key to an object was being changed.

If you feel strongly about it, maybe I could use a dedicated class instead of forcing the behavior on primitives? i'll take a look at it.

app/assets/javascripts/angular/app.js
```diff
@@ -15,4 +15,6 @@ angular.module("Prometheus",
   $rootScope.alert = function(thing) {
     alert(thing);
   };
+}]).config(['$locationProvider', function($locationProvider) {
+   $locationProvider.html5Mode(true);
```
stuartnelson3: bad news. this is required to make the custom pathing to work (`promdash.com/dashboard/profile`), but changing to html5 mode was shot down in this pr by @juliusv: https://github.com/prometheus/promdash/pull/422

grobie:
## legend items

I think it's fine to assume the graphs are similar. How are the legend settings stored? Let's say there are three sets of legends: `profile1: [a, b, c]; profile2: [a, b, d], profile3: [a, b]`. If I don't want to show `c` then it would be hidden in profile1 and simply ignored in profile2 and profile 3. Would htat work? The only real issue comes when someone hides `a, b` and then switches to profile3. As long as this doesn't raise an error, I think it would be fine to show an empty graph with all items disabled.

## profile names / slugs

I meant that we could automatically create a slug for name, removing all characters which are invalid in an URL. My alternative suggestion was to reject any profile name which doesn't match the regexp, so that no invalid characters can be entered. The third option would be to leave it as it is and have ugly URLs.

## template variables

Let's just add a note/warning for now in template variables. We can deal with that later.

stuartnelson3:
- profile names / slugs: happy to keep it ugly :p 
- template variables: note added
- legend items: hide a legend item in profile 1. switch to profile 2. hidden legend item is not persisted. switch back to profile 1. legend item is hidden again. modifications to legends are maintained on a per-profile basis owing to how legend items are stored, because we use unique keys to store legend item state.

### how are legend items stored?

for each legend, if the legend for an item is clicked, then all disabled items (an attribute that is toggled on/off by interacting with the legend) are added to a javascript object which is stored in memory (it is also persisted when a dashboard is saved)

these items are identified by their "unique name", or, their full expression name:
```
some_expression{instance="ip-1-2-3-4",label2="bleh",label3="other thing"}
```

so, the issue with this is that the unique names used to persist legend settings are (unsurprisingly) unique. from one profile to the next, there is no way of knowing what the original query was without doing some real hacky reverse variable interpolation (don't even think about it).

**because of the stored expression identifiers being unique, I can't tell which series in other profiles should be similarly hidden.**

1. update the list of disabled elements on click https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/graph_chart.js#L33
2. store an expression by it's unique name in the "disabled" hash if it should be https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/graph_chart.js#L462
3. on graph render, set an object as disabled if it's found in the object https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/graph_chart.js#L268-L276

the current implementation, as you see above, clears the object that stores the disabled series when updated. so if you make changes to the legend in profile 1, then make changes to the legend in profile 2, it wipes out the changes you had in profile 1.

it's easy to change this so you can store changes across all profiles:

```javascript
function storeLegendState() {
  if (!rsGraph) {
    return;
  }

  rsGraph.series.forEach(function(s) {
    if (s.disabled) {
      scope.graphSettings.disabledSeries[s.uniqName] = true;
      return
    }

    delete scope.graphSettings.disabledSeries[s.uniqName];
  });
}

```

is that good?

grobie:
Thanks for the explanation. Coming up with an identifier making the request possible seems to be complicated indeed. It makes sense to not work on that, please just make it possible to store legend settings in each profile separately.

stuartnelson3:
pushed

grobie:
Aaaaaaaaaaawesome!!! Great work man, this will make dealing with regions and other templates so much easier!

:+1:

PR #403, opened 2015-05-20
Only set logarithm-based graph min if logarithm scales are present.

stuartnelson3:
Setting the scale when there are series with both positive and negative
values according to a logarithmic minimum causes the minimum of the
graph to drop way out of control. This is only set when a logarithmic
scale exists.


juliusv:
:+1: When testing this I found all kinds of other oddities about the scales, but those are separate :) (like 2nd axis number ticks not being shown at all - is that known?)

PR #408, opened 2015-06-05
Native graphite support

stuartnelson3:
This pr adds native rendering of graphite graphs.

1. Create your graphite server in the servers tab. Make sure to check "this is a graphite server".
2. Create your widget as normal, entering your target in the query box.
3. profit

nb you can render graphite and prometheus timeseries on the same widget now

@matthiasr @juliusv @grobie @bjk-soundcloud


app/views/servers/_form.html.erb
```diff
@@ -19,6 +19,11 @@
     <%= f.label :url %><br>
     <%= f.text_field :url, class: "form-control" %>
   </div>
+  <div class="field form-group">
+    <label>
+    <%= f.check_box :kind, {}, "graphite", "prometheus" %>Graphite server
```
matthiasr: I think it would be prettier if this were a dropdown ("Prometheus", "Graphite", with the former being the default). But check box works for now too

matthiasr:
Whooohoooo :+1:

stuartnelson3:
can you give it another try? entering a metric in graphite should now show a list of suggestions

matthiasr:
I'm not a fan of the *-based autocompletion. It has the same issue as the graphite composer, it makes it nigh impossible to actually enter a wildcard.

stuartnelson3:
that's fair. @juliusv @grobie thoughts?

juliusv:
The select dropdown looks funny:

select

Could you make it look Bootstrap-style? :)

app/helpers/application_helper.rb
```diff
@@ -18,6 +18,14 @@ def dashboards_active?
     %w{dashboards directories}.include? params[:controller]
   end
 
+  def supported_backends
+    b = ["prometheus", "graphite"]
```
juliusv: At least in the human-visible representation, these should be capitalized.

app/assets/javascripts/angular/services/rickshaw_data_transformer.js
```diff
@@ -23,31 +23,29 @@ angular.module("Prometheus.services").factory('RickshawDataTransformer', [functi
...
-    series = new Rickshaw.Series(series);
+    series = series.concat((data.data.Value || data.data.value).map(function(ts) {
+      var name = metricToTsName(ts.Metric || ts.metric);
+      return {
+        name: name,
+        // uniqName is added to be kept as a unique, unmodified identifier for a series.
+        uniqName: name,
+        type: data.type,
+        axisID: axisIDByExprID[data.exp_id],
+        exp_id: data.exp_id,
```
juliusv: Ewww. We should really fix the mix of under_score names and camelCase where possible.

app/views/servers/_form.html.erb
```diff
@@ -19,6 +19,10 @@
     <%= f.label :url %><br>
     <%= f.text_field :url, class: "form-control" %>
   </div>
+  <div class="field form-group">
+    <%= f.label :backend %><br>
```
juliusv: The whole page is about defining a backend server. I think in this context, it makes more sense to call this "Server type" rather than "Backend".

app/views/servers/show.html.erb
```diff
@@ -9,6 +9,11 @@
   <%= @server.url %>
 </p>
 
+<p>
+  <strong>Kind:</strong>
```
juliusv: The naming here would be good to make consistent ("Kind" here and "Backend" elsewhere). How about calling it "Type" or "Server type" everywhere"?

stuartnelson3:
@juliusv thanks for the review, I addressed all your comments

what do you think about the auto-complete for graphite queries?

juliusv:
Hmm, Graphite-style metrics indeed suck for autocompletion, unless you already know that you're at a filtering stage where only few enough possibilities exist. I don't know - maybe no autocompletion for Graphite metrics?

stuartnelson3:
k, i'll go ahead and remove them

juliusv:
Unless @matthiasr has a better idea...

stuartnelson3:
I'll wait for a consensus. it's just a matter of leaving or reverting 17525b7287ba695e6309ffc07e9a8aa8890d8e31

juliusv:
Talked to @bjk-soundcloud and @matthiasr.

It would be better if at every stage, it would immediately show all autocomplete options for that dotted component, without having to type a star. A star should always be interpreted as "yes, I really want a star" here.

Disregard what I initially said about autocomplete sucking in general for Graphite. I didn't notice you were doing the autocomplete limited to one dot-separated component at a time instead of for the whole remaining metric name. So that should be fine.

stuartnelson3:
can that be done in the next PR?

juliusv:
Yup.

juliusv:
:+1: otherwise.

PR #414, opened 2015-06-10
Hide scrollbar on Linux chrome+firefox

grobie:
Please check @stuartnelson3. I don't have OSX, works fine on Ubuntu+Chromium.


stuartnelson3:
looks good :+1:

PR #418, opened 2015-06-15
fix hostname filter

### LINKED ISSUE #415: Regex parse labels?
Is it possible (or could it be made possible) to format legend labels with regex's applied to the actual time series? Mostly so long instance-name's can be filtered down to the small chunk of relevant data for display.

i.e. server.domain.com:9100 => server

stuartnelson3:
hostname filter was broken if a scheme was not present.

uses a  modified version of the regex found at
http://www.ietf.org/rfc/rfc3986.txt appendix B

fixes #415

@grobie @juliusv


stuartnelson3:
ignore

PR #422, opened 2015-06-17
Turn on html5 mode.

stuartnelson3:
A bug was found wherein a user was unable to render a promdash dashboard
within a frame widget if that dashboard relied on query string
parameters.

Without html5 mode engaged, the query string was only interpreted by the
angular code if it was preceded by a hash (`#`), e.g.:

`http://promdash.com/my-dashboard/#/?some=components&and=others`

However, the URL escaping code executed on frame URLs escapes all code
after the final hash (`#`), would would escape the `?` and break the
dashboard rendered in the frame.

A work-around for now is to add another `#` and some garbage behind it:

`http://promdash.com/my-dashboard/#/?some=components&and=others#garbageHash`

With html5 mode engaged, the above URL can be written as one is
accustomed now:

`http://promdash.com/my-dashboard?some=components&and=others`

__NOTE:__

This will break the variable parsing in current dashboards relying on the `/#?` format. Removing the `#` causes the variables to be parsed correctly.

@juliusv @grobie

### DIFF
--- a/app/assets/javascripts/angular/app.js
+++ b/app/assets/javascripts/angular/app.js
@@ -15,4 +15,6 @@ angular.module("Prometheus",
   $rootScope.alert = function(thing) {
     alert(thing);
   };
+}]).config(['$locationProvider', function($locationProvider) {
+  $locationProvider.html5Mode(true);
 }]);
--- a/app/assets/javascripts/angular/app.js
+++ b/app/assets/javascripts/angular/app.js
@@ -6,7 +6,7 @@ angular.module("Prometheus.filters", []);
 
 angular.module("Prometheus",
   ["ui.sortable", "ui.bootstrap", "ui.slider", "Prometheus.controllers", "Prometheus.directives", "Prometheus.resources", "Prometheus.services", "Prometheus.filters"])
-.run(['$rootScope', function($rootScope) {
+.run(['$rootScope', '$location', function($rootScope, $location) {
   // adds some basic utilities to the $rootScope for debugging purposes
   $rootScope.log = function(thing) {
     console.log(thing);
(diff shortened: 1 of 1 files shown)


juliusv:
Is it at all possible/realistic to change the escaping done on frame URLs instead? I'm quite wary of breaking all the existing links. Not only do they break, but other behavior also subtly breaks: for example, when I modify the time in the URL with the new code (and press Return), it will reload the entire page, whereas previously it correctly updated the time without reloading the dashboard. This has implications for losing unsaved changes, etc.

stuartnelson3:
oh goofy, that should be caught by angular and prevented (or so i thought).

i agree, that is exceedingly lame.

unfortunately i can't think of an immediate way to fix this. how could we differentiate between a frame link that has a `?` after it and needs to be escaped and a frame link that is formed like one of promdash's?

stuartnelson3:
for context, the escaping was added to deal with graphite URLs that have `colorlist=#aaaaaa,#bbbbbb,#cccccc`. Graphite can now be natively rendered, so maybe the hash escaping isn't necessary?

juliusv:
Setting explicit colors per time series isn't possible in Rickshaw graphs yet, so this would actually be the most important legacy-graphite graph feature we'll want to continue to support.

grobie:
:+1: on switching to HTML5 mode, as it seems to be a requirement for more and more features. The current angular format is highly confusing and not well understood by most users.

I do agree though that we should not break existing links. Is it possible to redirect from the current `#!` format to html5? I don't think that "I want to modify the URL without reloading" is an important feature to keep. There are UI elements available to change parameters without touching the URL. This should not be part of our interface.

stuartnelson3:
@grobie @juliusv try copying some links from our prod promdash, with `#?`, and paste them a browser running this code (but connected to the prod db). when i tested it, it appeared to be correctly changing the link and loading the data in the query string.

grobie:
:+1:

grobie:
Awesome! Works fine in all my tests of existing links. Let's go forward here.

PR #423, opened 2015-06-17
Consolidate graphite name handling

grobie:
@stuartnelson3


grobie:
@stuartnelson3 :+1: in RL.

PR #426, opened 2015-06-21
Add a cache for annotations requests.

stuartnelson3:
Annotations requests are cached based on the dashboard's endTime and
range. If the dashboard has no set endTime (i.e. the annotation endTime
requested will keep changing), the cache for that range+endTime
combination is erased every 5 minutes.

Previously, annotations were requested every time for each individual dashboard.

@juliusv @grobie

### DIFF
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -6,15 +6,13 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
                                                       "YAxisUtilities",
                                                       "SharedWidgetSetup",
                                                       "Palettes",
-                                                      "AnnotationRefresher",
                                                       function($scope,
                                                                $http,
                                                                $timeout,
                                                                GraphRefresher,
                                                                YAxisUtilities,
                                                                SharedWidgetSetup,
-                                                               Palettes,
-                                                               AnnotationRefresher) {
+                                                               Palettes) {
   SharedWidgetSetup($scope);
 
   // TODO: Set these on graph creation so we don't have to keep doing these
@@ -135,7 +133,6 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       debounce = $timeout(function() {
         refreshFn($scope.graph.endTime, $scope.graph.range, step).then(function(data) {
           scope.$broadcast('redrawGraphs', data);
-          AnnotationRefresher(scope.graph, scope);
         });
       }, timeout);
     };
@@ -154,8 +151,6 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
   $scope.$watch('graph.resolution', function() {
     $scope.refreshGraph();
   });
-
-  $scope.refreshGraph();
 
   if (location.pathname.match(/^\/w\//)) { // On a widget page.
     $scope.widgetPage = true;
--- a/app/assets/javascripts/angular/services/annotation_refresher.js.erb
+++ b/app/assets/javascripts/angular/services/annotation_refresher.js.erb
@@ -1,31 +1,66 @@
-angular.module("Prometheus.services").factory('AnnotationRefresher', ["$http", "VariableInterpolator", function($http, VariableInterpolator) {
-  return function(graph, scope) {
-    var tags = graph.tags.map(function(e) {
-      if (e.name) {
-        var n = VariableInterpolator(e.name, scope.vars);
-        return n.split(",").map(function(s) { return s.trim(); });
-      } else {
-        return "";
-      }
-    });
-    if (tags.length) {
-      var range = Prometheus.Graph.parseDuration(graph.range);
-      var until = Math.floor(graph.endTime || Date.now()) / 1000;
-
-      tags.forEach(function(t) {
-        $http.get('<%= Rails.configuration.path_prefix %>annotations', {
-          params: {
-            'tags[]': t,
-            until: until,
-            range: range
-          }
-        })
-        .then(function(payload) {
-          scope.$broadcast('annotateGraph', payload.data.posts);
-        }, function(response) {
-          scope.errorMessages.push("Error " + response.status + ": Error occurred fetching annotations.");
-        });
+angular.module("Prometheus.services").factory('AnnotationRefresher', ["$http", "$q", "$timeout", "VariableInterpolator", function($http, $q, $timeout, VariableInterpolator) {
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/angular/services/annotation_refresher.js.erb
```diff
@@ -1,31 +1,67 @@
...
+          endTime = "now";
+        }
+        var k = "" + range + endTime;
+        if (!promises[k]) {
+
+          if (endTime === "now") {
+            // If endTime isn't set, a long-running dashboard with
+            // auto-refresh will continually use annotations from the initial
+            // page load. This key needs to be deleted periodically.
+            // TODO: 5 min is arbitrary. Should it be something else?
```
stuartnelson3: Thoughts?

app/assets/javascripts/angular/services/annotation_refresher.js.erb
```diff
@@ -1,31 +1,67 @@
...
-        })
-        .then(function(payload) {
-          scope.$broadcast('annotateGraph', payload.data.posts);
-        }, function(response) {
-          scope.errorMessages.push("Error " + response.status + ": Error occurred fetching annotations.");
-        });
-      });
+
+          promises[k] = tags.map(function(t) {
+            return $http.get('<%= Rails.configuration.path_prefix %>annotations', {
```
stuartnelson3: A question that is out of scope for this pr: should we handle configuration by querying an endpoint server-side? Making this a `.js.erb` file means I can't unit test this with jasmine, as it needs to be run through ERB first.

app/assets/javascripts/angular/services/annotation_refresher.js.erb
juliusv: If we mainly want to debounce a single whole dashboard refresh, 1 minute should be sufficient here?

app/assets/javascripts/angular/services/annotation_refresher.js.erb
juliusv: How about injecting that into the page as a global variable, like we do for the dashboard configs and the servers?

app/assets/javascripts/angular/services/annotation_refresher.js.erb
```diff
@@ -1,31 +1,73 @@
...
-  return function(graph, scope) {
-    var tags = graph.tags.map(function(e) {
-      if (e.name) {
-        var n = VariableInterpolator(e.name, scope.vars);
-        return n.split(",").map(function(s) { return s.trim(); });
-      } else {
-        return "";
+angular.module("Prometheus.services").factory('AnnotationRefresher', ["$http", "$q", "$timeout", "VariableInterpolator", function($http, $q, $timeout, VariableInterpolator) {
+  var MINUTE = 1000*60;
+  var promises = {};
```
juliusv: I don't see you ever cleaning out this hash unless someone changes the tags? (but not for changes just to the range / endtime, except for the `now` case)?

stuartnelson3: Yeah, I'm never cleaning it out. I could put them all on individual 1 min cooldowns? It may be an incorrect assumption, but I'm guessing that any long running dashboards see very little churn in their tags or endtime/range, so I just assumed the `now` would be the only case in which data would be getting refreshed.

guess there aren't any downsides to making the request every minute.

app/assets/javascripts/angular/services/annotation_refresher.js.erb
stuartnelson3: hm k. do we want to do that here, or in a separate PR and wipe them all out at once?

stuartnelson3:
@juliusv hey, made it so all queries are cleared from the cache after 1min, and moved `Rails.configuration.path_prefix` to a global

app/assets/javascripts/angular/services/annotation_refresher.js.erb
grobie: What speaks against parsing all templates before running jasmine? You might not even need a full rails app for that by only binding all required variables (e.g. `Rails.configuration.path_prefix`).

app/assets/javascripts/angular/controllers/graph_ctrl.js
```diff
@@ -135,7 +135,6 @@ angular.module("Prometheus.controllers").controller('GraphCtrl',
       debounce = $timeout(function() {
         refreshFn($scope.graph.endTime, $scope.graph.range, step).then(function(data) {
           scope.$broadcast('redrawGraphs', data);
-          AnnotationRefresher(scope.graph, scope);
```
grobie: Looks like you can remove `AnnotationRefresher` from the module dependencies now.

app/assets/javascripts/angular/services/annotation_refresher.js.erb
grobie: What is the reason you want to clean up the cache. Reducing memory footprint / preventing memory leaks? In that case it would make more sense to ensure the the lookup dictionary can't contain more than x elements, by throwing out the oldest entry in case the dictionary is full. Currently it's still possible to add a lot of entries (plus timeout() functions) by changing the times a lot.

The cleanup timeout of 1 minute also feels quite arbitrary. Why not using the refresh interval instead?

app/assets/javascripts/angular/services/annotation_refresher.js.erb
stuartnelson3: I'm assuming we can just parse the templates in [`karma.rake`](https://github.com/prometheus/promdash/blob/master/lib/tasks/karma.rake), but that seems like making more work for ourselves instead of just removing the rails templating from our js files

app/assets/javascripts/angular/services/annotation_refresher.js.erb
stuartnelson3: dictionaries are unordered, so we would have to add some additional logic to manage which is the oldest.

The main purpose of this PR was to only make a single request. We could just delete the shared annotation server response as soon as it has been consumed, by putting it on a timer to clear after some (again, arbitrary) small amount of time within which all graphs should have finished loading, e.g. 10sec.

juliusv: Sounds good, ok make it 1 minute.

juliusv: Ah you already did that. Ignore.

grobie:
:+1:

PR #428, opened 2015-06-22
Code assumes that there is a trailing slash behind all server URLs.

### LINKED ISSUE #425: "Link to Graph in Prometheus" only works when server url was configured with trailing /

stuartnelson3:
Enforce this slash's existence on save for server.

fixes #425 

@juliusv

### DIFF
--- a/app/assets/javascripts/angular/directives/expression.js.erb
+++ b/app/assets/javascripts/angular/directives/expression.js.erb
@@ -2,10 +2,12 @@ angular.module("Prometheus.directives").directive('expression', ["$http",
                                                   "$timeout", "$filter",
                                                   "MetricNamesQuerier",
                                                   "VariableInterpolator",
+                                                  "URLGenerator",
                                                   function($http, $timeout,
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -205,4 +205,4 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
 
 BUNDLED WITH
-   1.10.1
+   1.10.4
```
stuartnelson3: i can't believe this is part of the new bundler -.-

juliusv:
This would add a second mechanism for slash sanitization. In all other places, we already use the `URLGenerator` to ignore the exact suffix. Can't we do the same for fixing https://github.com/prometheus/promdash/issues/425?

Gemfile.lock
juliusv: Yeah. Sheesh...

stuartnelson3:
that's way better, updated

juliusv:
:+1:

PR #432, opened 2015-07-22
Gauge fixes

### LINKED ISSUE #420: Gauge doesn't appear to support template variables in data source
I tried using "{{my_var}}" as a data source. I get this error:

    Expression {{my_var}}: Parse error at char 2: unexpected left brace '{'

This seems to work for Graphs, so I think this might be something that's only broken for Gauges.

stuartnelson3:
@juliusv in the process of adding the variable interpolator, i also noticed that any attempt to show a 0 value for the gauge would not graph because 0 is falsy in js

fixes #420


juliusv:
Hmm, I can enter `1` and it gets shown fine, but `0` still doesn't do anything even with this PR?

juliusv:
:+1:

PR #433, opened 2015-07-24
Bumped eventmachine and mysql2 gem versions

y3ti:
Eventmachine 1.0.3 and mysql2 0.3.14 don't compile with ruby 2.2


juliusv:
:+1: Thanks!

Issue #434, opened 2015-08-07
Setting PROMDASH_PATH_PREFIX does not work for assets


mingfang:
When I set PROMDASH_PATH_PREFIX, eg ```promdash```, the dashboard comes up but all the assets are broken.

I don't know Ruby but I was able to hack it to work by removing line 40 from https://github.com/prometheus/promdash/blob/master/config/application.rb

juliusv:
So it also doesn't work for me. The assets are served under the correct prefix, but the links to them are generated without the prefix. Even links that are clearly generated through Rails helpers are wrong: `asset_path('graph_template.html')` in https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/graph.js.erb#L11.

I found http://serverfault.com/questions/638905/does-rails-4-asset-path-helper-uses-asset-prefix. Regarding the example I mentioned above, the actual graph HTML template file has an `.html.erb` extension instead of just `.html`. Referring to just `.html` works without a path prefix, but breaks when setting one. Adding the missing `.erb` unbreaks the loading of said graph template file.

Poking a bit further...

juliusv:
@mingfang Could you take a look if https://github.com/prometheus/promdash/pull/438 fixes your problem?

mingfang:
It’s still not working for me.

I’m getting 404s on stylesheets/application.css  and javascripts/base.js.

juliusv:
@mingfang Hmm, that's strange. How are you running PromDash (Docker, non-Docker, development or production Rails env?).

I just tested my PR branch from a completely clean checkout and in an incognito browser window, and everything loads correctly (with `PROMDASH_PATH_PREFIX=/prefix`).

juliusv:
I did:

```
git clone git@github.com:prometheus/promdash
git checkout origin/fix-path-prefix
bundle
bundle exec rake db:migrate
DATABASE_URL="<my-db-url>" PROMDASH_PATH_PREFIX=/prefix bundle exec rails s
```

mingfang:
I’m running it using my own Dockerfile.

Please see lines 31-36 here 
https://github.com/mingfang/docker-prometheus/blob/master/Dockerfile

The Runit script is here
https://github.com/mingfang/docker-prometheus/blob/master/sv/promdash/run

I’m going to try your method to see if that works.

juliusv:
That'd be great. I tried the Dockerfile, but it needs the external files that are `ADD`-ed.

[...]

PR #438, opened 2015-08-10
Fix asset URLs when using path prefix.

juliusv:
This fixes https://github.com/prometheus/promdash/issues/434

### DIFF
--- a/app/assets/javascripts/angular/directives/gauge.js.erb
+++ b/app/assets/javascripts/angular/directives/gauge.js.erb
@@ -8,6 +8,6 @@ angular.module("Prometheus.directives").directive('gauge', function() {
       vars: "="
     },
     restrict: "AE",
-    templateUrl: "<%= asset_path('gauge_template.html') %>"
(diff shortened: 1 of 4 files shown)


juliusv:
@stuartnelson3 @quinox @mingfang

app/assets/javascripts/angular/directives/gauge.js.erb
```diff
@@ -8,6 +8,6 @@ angular.module("Prometheus.directives").directive('gauge', function() {
       vars: "="
     },
     restrict: "AE",
-    templateUrl: "<%= asset_path('gauge_template.html') %>"
+    templateUrl: "<%= asset_path('gauge_template.html.erb') %>"
```
grobie: Hmmm. Do we really want to link to an erb file?

juliusv:
There were other fix PRs around this already now, but now with precompiled assets in production env, it doesn't even load the application.{js,css} correctly anymore when using a path prefix :( Gah, Rails. This needs another closer look.

PR #439, opened 2015-08-10
Fix the way that Bootstrap errors are rendered

samstarling:
* In Bootstrap 3, the `-danger` suffix is used, not `-error`.
* The name parameter is a string, so the symbol comparison was failing.

:wave: to @juliusv and @stuartnelson3.


grobie:
:+1:

juliusv:
:+1: Thanks!

PR #442, opened 2015-08-14
Create unique slug for dashboards

stuartnelson3:
cases:
- dashboard slug already exists
- dashboard attempts to use blacklisted slug name

the behavior with blacklisted slugs was unexpected, but it has the result of letting users create dashboards with names that would normally have been blacklisted, but now are assigned a unique name.

@grobie

### DIFF
--- a/app/models/dashboard.rb
+++ b/app/models/dashboard.rb
@@ -16,7 +16,7 @@ class Dashboard < ActiveRecord::Base
   belongs_to :directory
   has_many :shortened_urls
 
-  validates :name, uniqueness: { case_sensitive: false }
+  validates :name, :slug, uniqueness: { case_sensitive: false }
   validates :name, :slug, presence: true
   validate :acceptable_slug
   validates :slug,
@@ -32,7 +32,7 @@ class Dashboard < ActiveRecord::Base
 
   def self.new_with_slug(params)
     dashboard = new(params)
-    dashboard.create_slug
+    dashboard.create_slug dashboard.name
     dashboard
   end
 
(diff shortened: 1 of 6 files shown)


app/models/dashboard.rb
```diff
@@ -64,7 +64,10 @@ def widgets
     (JSON.parse dashboard_json)['widgets']
   end
 
-  def create_slug
-    self.slug = SlugMaker.slug(name)
+  def create_slug name, n=nil
+    self.slug = SlugMaker.slug("#{name} #{n}")
+    if !valid? && errors[:slug].any?
```
stuartnelson3: errors hash, as far as i could tell, isn't generated until `valid?` is called

app/models/dashboard.rb
```diff
@@ -16,7 +16,7 @@ class Dashboard < ActiveRecord::Base
   belongs_to :directory
   has_many :shortened_urls
 
-  validates :name, uniqueness: { case_sensitive: false }
+  validates :name, :slug, uniqueness: { case_sensitive: false }
```
grobie: While this is great first step to validate uniqueness, the only way to really ensure it is to have a unique constraint on the database. Otherwise it's possible that in case of race conditions (two requests get executed at the same time) we still end up with duplicates.

app/models/dashboard.rb
grobie: Yeah. You could also check explicitly test whether there is a slug with that name already. Might have less side effects: `self.class.exists?(slug: slug)` or so.

app/models/dashboard.rb
stuartnelson3: always the stickler. added.

grobie:
:+1:

PR #445, opened 2015-08-27
Minor UI changes to the Dashboard

samstarling:
The visual changes here are fairly minor, but there are a few differences under the hood – I'll list those below.

<h4>Before</h4>
before

<h4>After</h4>
after

#### Changes

* Per-graph popups now have slightly cleaner formatting: inputs and controls use more of the available space.
* Tidied up some of the spacing between elements to be more uniform, same goes for the fullscreen view.
* Slightly improved use of Bootstrap classes like `row` and `col-...`.
* Some whitespace changes - sorry for not keeping these separate.

Please let me know if you'd like me to squash/tidy these commits. I left them separate for now, in case you take particular issue with any of them individually. There's probably more where this came from, but I thought I'd keep it simple for now.

Thanks!

/cc @juliusv @stuartnelson3 @grobie

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.css.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.css.scss
@@ -3,4 +3,5 @@ body { padding-top: 60px; }
 
 .panel-heading {
   font-weight: bold;
+  padding: 8px 10px 10px 10px;
 }
--- a/app/assets/stylesheets/dashboards.css.scss
+++ b/app/assets/stylesheets/dashboards.css.scss
@@ -69,6 +69,7 @@ $light-secondary-background: #f0f0f0;
 .widget_wrapper {
   background-color: $light-secondary-background;
   border: none;
+  padding: 5px;
   width: 100%;
 }
 
--- a/app/assets/stylesheets/dashboards.css.scss
+++ b/app/assets/stylesheets/dashboards.css.scss
@@ -39,10 +39,13 @@ $light-secondary-background: #f0f0f0;
 }
 
 .fullscreen {
-  position: absolute;
+  position: fixed;
   top: 0;
   left: 0;
-  width: 100%;
+  right: 0;
+  height: 100%;
+  overflow-y: scroll;
+  overflow-x: hidden;
   z-index: 1050; // Bootstrap's navbar is at 1030.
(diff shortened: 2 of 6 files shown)


samstarling:
FYI: I haven't run the tests yet!

stuartnelson3:
have you checked visually between safari/chrome/firefox?

samstarling:
Still to do:

[ ] Check in Safari, Chrome and Firefox
[ ] Run the tests

samstarling:
Looks good in Firefox and Safari.

juliusv:
@samstarling Thanks! Taking a closer look now.

I noticed this on the index (directories) page:

dir_page

The action icons look different / broken...

juliusv:
Hmm, same for the actions on a graph, but not sure yet why exactly.

juliusv:
The problem went away for me with a fresh checkout of PromDash. Apologies.

juliusv:
Getting this weird extra element underneath the expression in the Datasources tab of the graph now:

ui_glitch

Looking at its DOM, it seems to be related to the typeahead autocomplete plugin...

stuartnelson3:
@juliusv @samstarling the problem with the typeahead is due to it being upgraded in a previous PR. Looking at the source code, the upgraded version of the typeahead assumes a version of angular higher than what we are using in promdash.

a previous attempt at upgrading angular was abandoned after we encountered weird, linux-only errors.

the easiest solution would be to try to upgrade our copy of angular again.

the more difficult solution is either to edit our vendored copy of ui-bootstrap, or roll back to a previous version.

juliusv:
I noticed this PR solves the scrollbar part of https://github.com/prometheus/promdash/issues/453.

Since the autocomplete problem exists with or without this PR, I'll merge this to fix at least the scrollbar.

Thanks!

samstarling:
@juliusv Thanks! If there's anything here that looks like it's broken something, I tried to make the commits small and atomic, so hopefully you'll be able to pick out the "bad" commit if things break. I did also notice it had the side effect of making the drag-and-drop rearrangement a little smoother. Give me a shout if things look broken, I'm aware I didn't do too well with this last time :smile:

juliusv:
@samstarling No problem, thanks for this :)

PR #448, opened 2015-09-08
fixing icons in RAILS_ENV="production"

ashmere:
correct the default location of bootstrap fonts to the /assets directory other assets use.
enable the fallback compile of missing assets allowing icons to work in production mode again 

With the change to rails 4 the asset pipeline changed from what I
understand in rails 3 assets were generated with both an md5 digest and
normal name, in rails 4 only the md5 digest is created
see
https://devcenter.heroku.com/articles/rails-4-asset-pipeline
a full fix should include changes to include asset_path.


juliusv:
Thanks, giving this whole problem a closer look now to understand what's broken and what the best thing is to do. FTR, PromDash has always been Rails 4, but I guess this problem was introduced with the latest Bootstrap upgrade?

/cc @samstarling

juliusv:
Ok, so the problem happened because the old bootstrap had some custom changes to make the font paths work (replaced `url()` with `image-url()` calls). The problem was, those overrides were placed directly into the `bootstrap.css`, not in the overrides CSS file, so they got lost during the bootstrap upgrade. Doh!

If you put the following into `bootstrap_and_overrides.css.scss`, the asset compilation fallback change in `production.rb` should no longer be necessary (at least it works for me in production mode without setting `config.assets.compile = true`):

```css
@font-face {
  font-family: 'Glyphicons Halflings';

  src: image-url('glyphicons-halflings-regular.eot');
  src: image-url('glyphicons-halflings-regular.eot?#iefix') format('embedded-opentype'), image-url('glyphicons-halflings-regular.woff2') format('woff2'), image-url('glyphicons-halflings-regular.woff') format('woff'), image-url('glyphicons-halflings-regular.ttf') format('truetype'), image-url('glyphicons-halflings-regular.svg#glyphicons_halflingsregular') format('svg');
}
```

ashmere:
followed your suggestion an updated the PR,
This now also includes the version change to 3.3.5 from 3.0.0 bootstrap it looks like this was supposed to be done in a previous comment but files were not included ??

juliusv:
Thanks! Indeed, half of the files were forgotten in the boostrap upgrade (https://github.com/prometheus/promdash/commit/211272c0d8592be0e375a3dab755706dee478d09).

Could you squash your commits either into one, or if you prefer, into two (one adding the missing bootstrap files, one adding the necessary overrides).

ashmere:
@juliusv single commit now

juliusv:
Thanks a bunch! Merging...

PR #449, opened 2015-09-13
Documentation updates for dev tools and a simple reverse proxy deployment

jonnenauha:
I commented on a issue that this aspect of prometheus and promdash is not covered very well. I'm hoping new prometheus users like me will benefit from it written down somewhere. My biggest hurdle in even trying out prometheus for a long time was the lack of clear docs for how to restrict access to the monitoring.

As @juliusv noted in the issue the projects focus has not been on this kind of "small scale" setup you roll on your own, but for more experienced people and private clusters. I believe many small companies might benefit in what I wrote down as its fairly simple to setup. This is for small deployments as everything is running on a single machine and means you cant easily add more prometheus servers. You can by doing more reverse proxy magic and esp. once the basic auth in the server URL is implemented.

I can also look into making a more in detail doc page into https://github.com/prometheus/docs with all details and configs for HAProxy as that is what I know how to use now. If you think its a good idea.

If the dev tool install part is too much detail in your opinion, I made it a separate commit so its easy to revert.

Feel free to comment on any typos or bad wording, I'll fix them up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,23 @@ $ jshint app/assets/javascripts
 ## Deployment
 
 ### Development Mode
+
+You need need to install multiple dependencies for the full development cycle.
+Here is the rundown for Ubuntu 14.04 LTS as a reference. If you run into errors
+with `bundle`, the most likely reason is that the database servers are not
+properly installed with dev headers/libs.
+
+* Node.js
+ * `sudo apt-get install nodejs nodejs-legacy npm`
+* Ruby
+ * `sudo apt-get install ruby-full` ([details](https://www.ruby-lang.org/en/documentation/installation/))
+* MySQL
+ * `sudo apt-get install libmysqlclient-dev` ([details](https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/index.html#apt-repo-fresh-install))
+* PostgreSQL
+ * `sudo apt-get install postgresql-9.{3|4} postgresql-server-dev-9.{3|4}` ([details](http://www.postgresql.org/download/linux/))
+* SQLite
+ * `sudo apt-get install sqlite3 libsqlite3-dev` ([details](https://www.sqlite.org/download.html))
+
 Install gems:
 
     bundle
@@ -65,10 +82,11 @@ Run the production server from the bundled environment:
     bin/env bin/bundle exec bin/thin -p $PORT start
 
 ### Deploy with Docker
+
 To deploy PromDash with Docker, use the [prom/promdash](https://registry.hub.docker.com/u/prom/promdash/) Docker image.
 By default, the image will start the [thin webserver](http://code.macournoyer.com/thin/)
 webserver in production mode. To run rake tasks like migrations, you
-can specify any kind of command as parameter to the Docker image.  
+can specify any kind of command as parameter to the Docker image.
 
 #### Super easy quickstart deployment with Docker
 
@@ -76,17 +94,69 @@ The following is a quick-start example for running PromDash with a file-based lo
 
 First, create the SQLite3 database in a shared local Docker volume on the host:
 
-    docker run -v /tmp/prom:/tmp/prom -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 prom/promdash ./bin/rake db:migrate
+    docker run --rm -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           prom/promdash ./bin/rake db:migrate
 
 Now, we launch PromDash with the database we've just created:
 
-    docker run -p 3000:3000 -v /tmp/prom:/tmp/prom -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 prom/promdash
+    docker run -p 3000:3000 -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           prom/promdash
+
+You can pass parameters directly to the thin server with:
+
+    docker run -p 3000:4000 -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           prom/promdash ./bin/thin -a localhost -p 4000 start
 
 ### Deploy behind a reverse proxy
 
 To deploy PromDash behind a reverse proxy you can set a global path prefix
 using the environment variable `PROMDASH_PATH_PREFIX`. Once set all URLs will
 start with the given prefix.
+
+One simple way to secure your Prometheus deployment with authentication is
+running a reverse proxy, Prometheus, and PromDash on the same machine. The goal
+is to have a publicly available deployment for authenticated users via your
+reverse proxy while restricting raw `IP:PORT` access. We won't cover any
+details on how to configure your reverse proxy, as it depends on what you are
+using.
+
+As PromDash executes Prometheus queries directly from the client it is simpler
+to use a single domain. This makes the server side ACL management simpler and
+the client requests already authenticated. 
+
+**Reverse proxy** (e.g. [nginx](http://nginx.org/), [HAProxy](http://www.haproxy.org/))
+* Set up single domain.
+ * Route `http(s)://dash.example.com/dash` to backend `localhost:3000`
+ * Route `http(s)://dash.example.com/prom` to backend `localhost:9090`
+* Configure the same ACL, e.g. HTTP basic auth for both backends.
+
+**Run Prometheus** with binding only to localhost to restrict direct external
+access over the network that would circumvent your authentication. Remember to
+pass the path portion in `-web.external-url`. Example for Docker:
+([full reference](http://prometheus.io/docs/introduction/install/#using-docker))
+
+    docker run -p 127.0.0.1:9090:9090 -v /prometheus-data \
+           prom/prometheus -config.file=/prometheus-data/prometheus.yml \
+                           -web.external-url "http://dash.example.com/prom"
+
+**Run PromDash** in a similar fashion. Once PromDash is running add a
+Prometheus server with `http://dash.example.com/prom/` in the web UI at
+`http://dash.example.com/dash`.
+
+    docker run -p 127.0.0.1:3000:3000 -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           -e PROMDASH_PATH_PREFIX=/dash \
+           prom/promdash
+
+Finally check that your reverse proxy is working as expected by visiting
+`dash.example.com/dash` and `dash.example.com/prom`. Both should prompt you for
+credentials, e.g. HTTP basic auth. You can use a Chrome incognito tab to force
+authentication. Next visit the public facing IP address on the ports
+`<IP>:3000/dash` and `<IP>:9090/prom` to verify they are in fact giving a page
+load error.
 
 ### Deployment Checklist
 


jonnenauha:
Hmm, running this documented setup once again in our cloud and does not seem to work again. dash cant still poll prom. Hold on the potential merge, I'll need to still sort this out :P Damn...

jonnenauha:
Ok, so I guess I've missed the idea of promdash by a big margin. Looking at the HAProxy logs the poll requests are actually coming from my IP not from the server. I was under the impression that prom polls the metrics, I cant say why but just assumed that would be more efficient :) (one querying instead of all clients who have the page open).

Alright, so this pull request cant be merged before the `http://user:pass@host:9090/` issue is resolved. After that I can do a slight tweak to this and it will actually be true :)

jonnenauha:
Now that I understood this I whitelisted our companys office IP in HAProxy and it started working. So this is enough for my use cases until that basic auth in the URL is implemented. If you think this would be good doc if I change it to note the whitelisting of "trusted srcs" I can update the docs too, and then update again with basic auth stuff.

juliusv:
Cool, thanks for doing this! More detailed comments coming now...

README.md
```diff
@@ -32,6 +32,20 @@ $ jshint app/assets/javascripts
 ## Deployment
 
 ### Development Mode
+
+You need multiple tools for the full development cycle. Here is the rundown for Ubuntu 14.04 LTS as a reference. If you run into errors with `bundle`, the most likely reason is that the database servers are not properly installed with dev headers/libs.
```
juliusv: Could you wrap all paragraphs to 80 chars, similar to the existing ones?

README.md
```diff
@@ -32,6 +32,20 @@ $ jshint app/assets/javascripts
...
+
+* nodejs
```
juliusv: Oh, I'm on that Ubuntu version, but don't have nodejs installed (`dpkg -l | grep node` -> empty output). Sure it's needed?

README.md
juliusv: "need multiple tools" -> "need to install multiple dependencies"

README.md
```diff
@@ -32,6 +32,20 @@ $ jshint app/assets/javascripts
...
+ * `sudo apt-get install nodejs nodejs-legacy npm`
+* ruby
+ * `sudo apt-get install ruby-full` ([details](https://www.ruby-lang.org/en/documentation/installation/))
+* mysql
+ * `sudo apt-get install libmysqlclient-dev` ([details](https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/index.html#apt-repo-fresh-install))
+* postgre
+ * `sudo apt-get install postgresql-9.{3|4} postgresql-server-dev-9.{3|4}` ([details](http://www.postgresql.org/download/linux/))
+* sqlite
```
juliusv: Let's use the proper names here: (Node.js), Ruby, MySQL, PostgreSQL, SQLite.

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
 using the environment variable `PROMDASH_PATH_PREFIX`. Once set all URLs will
 start with the given prefix.
 
+One simple way to secure your Prometheus deployment with authentication is running a reverse proxy, Prometheus and PromDash on the same machine. The goal is to have a publicly available deployment for authenticated users via your reverse proxy while restricting raw `IP:PORT` access. We wont go cover any details on how to configure your reverse proxy, it depends on what you are using.
```
juliusv: I'm an Oxford comma fan, especially since it can make this listing less ambiguous :)

So "reverse proxy, Prometheus, and PromDash".

juliusv: "wont go" -> "won't"

", it depends" -> ", as it depends"

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+
+**Reverse proxy** (eg. [nginx](http://nginx.org/), [HAProxy](http://www.haproxy.org/))
+* Setup two domains. Route requests by the specific hostnames and reject any other requests.
```
juliusv: "Setup" -> "Set up" (verb)

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+ * `http(s)://prometheus.example.com` to backend `localhost:9090`
+ * `http(s)://dash.example.com` to backend `localhost:3000`
+* Configure ACL eg. HTTP basic auth for both backends
```
juliusv: Comma after "ACL"

"eg." -> "e.g."

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+
+**Run Prometheus** with binding only to localhost to restrict raw `IP:PORT` access that would circumvent your auth. Remember to pass `-web.external-url`. Example for docker: ([full reference](http://prometheus.io/docs/introduction/install/#using-docker))
```
juliusv: The meaning of "raw" is not obvious. How about:

"Run Prometheus with binding only to `localhost` to restrict direct external access over the network that would circumvent your authentication."

juliusv: s/docker/Docker/

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+
+    docker run -p 127.0.0.1:9090:9090 -v /prometheus-data \
+           prom/prometheus -config.file=/prometheus-data/prometheus.yml \
+                           -web.external-url "http://prometheus.example.com/"
+
+**Running PromDash** with dockers default `--net brigde` won't let it poll Prometheus at `localhost:9090` from the host. We will need to use `--net host` to access it. This means we should again bind only to localhost to keep access restricted. Once PromDash is running add a Prometheus server with `http://localhost:9090/` in the web UI at `http://dash.examples.com`.
```
juliusv: dockers -> Docker's

README.md
```diff
@@ -88,6 +113,28 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+
+    docker run --net host -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           prom/promdash ./bin/thin -a localhost -p 3000 start
+
+Finally check your reverse proxy is working as expected by visiting `dash.examples.com` and `prometheus.example.com`. Both should promp you for credentials eg. HTTP basic auth. Next visit the public facing IP address in the ports `<IP>:9090` and `<IP>:3000` to verify they are in fact giving a page load error.
```
juliusv: promp -> prompt

juliusv: eg. -> e.g.

juliusv: Comma before e.g.

juliusv: "in the ports" -> "on the ports"?

README.md
jonnenauha: I'm fairly sure i needed to install it on the fresh VM. I think thats besides the point as this should be a reference in general for linux distros, even if happen to be using Ubuntu and `apt`. I think it should be explicitly stated that this tool is needed.

jonnenauha:
@juliusv thanks for the help. Hope I covered them all.

I'm thinking I should modify this a bit to actually document the single domain work around from https://github.com/prometheus/promdash/issues/341#issuecomment-74425398

I did this last night with HAProxy and it works well. I was able to expose `/prom/api > :9090/api` `/prom/graph > :9090/graph` and `/static > :9090/static`. That graph part is missing from the work around, it proved to be sometimes valuable to also have access to the graph at prometheus.

However this increases the complexity a bit and will make the examples longer and more complex to understand. I think I could just doc the /api/ for both nginx and haproxy?

This might be better than waiting for the basic auth for prom server endpoint fix. Would get something visible now, and if you want it could be changed to two domains later once the fix is in.

README.md
```diff
@@ -88,6 +116,44 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
 using the environment variable `PROMDASH_PATH_PREFIX`. Once set all URLs will
 start with the given prefix.
 
+One simple way to secure your Prometheus deployment with authentication is
+running a reverse proxy, Prometheus, and PromDash on the same machine. The goal
+is to have a publicly available deployment for authenticated users via your
+reverse proxy while restricting raw `IP:PORT` access. We won't go cover any
```
juliusv: "go cover" -> "cover"

README.md
```diff
@@ -88,6 +116,44 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+access over the network that would circumvent your authentication. Remember to
+pass `-web.external-url`. Example for Docker:
+([full reference](http://prometheus.io/docs/introduction/install/#using-docker))
+
+    docker run -p 127.0.0.1:9090:9090 -v /prometheus-data \
+           prom/prometheus -config.file=/prometheus-data/prometheus.yml \
+                           -web.external-url "http://prometheus.example.com/"
+
+**Running PromDash** with Docker's default `--net brigde` won't let it poll
+Prometheus at `localhost:9090` from the host. We will need to use `--net host`
```
juliusv: I didn't notice this earlier. As mentioned in https://github.com/prometheus/promdash/issues/341, PromDash does *not* need to talk to Prometheus at all. It only needs to be able to talk to its database (and perhaps an annotation server, if that's optionally configured, but let's ignore that here). Is the `--net host` setting still needed here then for other reasons?

juliusv:
Yep, I'd be all for describing the single-domain variant, as that's the only thing that is working at the moment (and nobody is actively working on a fix). Maybe just route all of `/prom` to Prometheus, not just certain whitelisted endpoints. It's all behind auth anyways, right?

README.md
jonnenauha: Yeah, this part will be rewritten and `--host net` is no longer needed.

jonnenauha:
Can you point me to to the right file/lines where I could take a peek at the basic auth problem. I have a lot of experience with js and have used angular in a few projects. Could try to figure out the basic auth bug.

`/prom` will work fine for the API. If you want `/graph` to render properly it will be a bit more complex. So I agree that the single prefix would be good.

I really wish prometheus would also allow to modify the root prefix that it mounts all its handlers to. I think this is what you do with `PROMDASH_PATH_PREFIX`.

If it did, you could just run it with `/prom` and no additional "hacks" with the reverse proxy would be required. Would fix the graph/static problem. I wonder if this could be done in the main repo, maybe I should make a issue there explaining the situation?

I try to update this PR in the upcoming days to something that works today.

stuartnelson3:
my idea regarding basic auth was written here: https://github.com/prometheus/promdash/issues/341#issuecomment-131769919

juliusv:
@jonnenauha Cool! Yeah, @stuartnelson3 already pointed to how to tackle the auth problem in general.

As for making Prometheus mount all its handlers on a path prefix, that already exists. But it's a command-line flag in the Prometheus server:

```
./prometheus  -h | grep -A3 -i external
   -web.external-url 
      The URL under which Prometheus is externally reachable (for 
      example, if Prometheus is served via a reverse proxy). Used for 
      generating relative and absolute links back to Prometheus itself. If 
      omitted, relevant URL components will be derived automatically.
```

jonnenauha:
Ok, surely read that but I failed to understand you can also give a path portion. How does that work exactly if the server is listening to `localhost:8080`. I suppose you parse the path part and prefix all handlers with it?

So would `-web.external-url "http://dash.example.com/prom"` result in it mounting root (index page) to `/prom/` and graph to `/prom/graph`?

This is what I was looking for.

juliusv:
Yeah, that's how it works - Prometheus will mount all its endpoints under `/prom` in your example, and will also adjust all links pointing to itself to include that path prefix.

jonnenauha:
Wow, thats great. Maybe you could update the doc string in the `-h` to reflect this better, something like 

> If the URL has a path portion, it will be used to prefix all the HTTP handlers.

You can probably work it more concisely but I had a hard time understand it does that from the current doc string, maybe I'm not the only one. Just a suggestion.

juliusv:
@jonnenauha Good point. The path prefix features botrh in Prometheus and PromDash were introduced by external contributors who needed them, so they haven't always gotten the full attention they deserved. Here's a PR for adding your suggestion: https://github.com/prometheus/prometheus/pull/1089

jonnenauha:
Alright, I've changed the documentation to use the single domain approach and tested in our production this actually work :) For good measure I added minimal HAProxy and Nginx (untested, modified from https://github.com/prometheus/promdash/issues/341#issuecomment-74425398) configs. 

If you think this is too much detail, takes too much space or is up to the user to figure out I made it a separate commit so its easy to revert.

juliusv:
Ok yeah, it's getting long, but I guess it's still ok to have it in the README.md for now. We might move it to a separate place at some point and link to it. Giving it a closer look now.

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
 using the environment variable `PROMDASH_PATH_PREFIX`. Once set all URLs will
 start with the given prefix.
 
+One simple way to secure your Prometheus deployment with authentication is
+running a reverse proxy, Prometheus, and PromDash on the same machine. The goal
+is to have a publicly available deployment for authenticated users via your
+reverse proxy while restricting raw `IP:PORT` access.
+
+As PromDash executes Prometheus queries directly from the client it is simpler
```
juliusv: Let's make super clear what "client" means by saying "browser" instead :) Like:

`As PromDash executes Prometheus queries directly via AJAX requests from the browser, ..."

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+to use a single domain. This makes the server side ACL management simpler and
+the client requests already authenticated.
```
juliusv: maybe "the client requests will use the same authentication for both PromDash and Prometheus."?

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+is to have a publicly available deployment for authenticated users via your
+reverse proxy while restricting raw `IP:PORT` access.
+
+As PromDash executes Prometheus queries directly from the client it is simpler
+to use a single domain. This makes the server side ACL management simpler and
+the client requests already authenticated. 
+
+**Reverse proxy** (e.g. [nginx](http://nginx.org/), [HAProxy](http://www.haproxy.org/))
+* Set up single domain.
+ * Route `http(s)://dash.example.com` to backend `localhost:3000`
```
juliusv: It'd be better to use a dedicated path prefix for PromDash too (otherwise, if someone names a PromDash dashboard `prom`, it would go to the wrong backend), as you've explained the usage of `PATH_PREFIX` above anyways...

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+
+**Run PromDash** in a similar fashion. Once PromDash is running add a
+Prometheus server with `http://dash.example.com/prom/` in the web UI at
+`http://dash.examples.com`.
+
+    docker run -p 127.0.0.1:3000:3000 -v /tmp/prom:/tmp/prom \
+           -e DATABASE_URL=sqlite3:/tmp/prom/file.sqlite3 \
+           prom/promdash
+
+Finally check your reverse proxy is working as expected by visiting
```
juliusv: "check" -> "check that" reads a bit better

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+`dash.examples.com` and `dash.example.com/prom`. Both should prompt you for
+credentials, e.g. HTTP basic auth. You can use a Chrome incognito tabs to force
```
juliusv: "a ... tabs" -> "a ... tab"

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+authentication. Next visit the public facing IP address on the ports
+`<IP>:3000` and `<IP>:9090` to verify they are in fact giving a page
+load error.
+
+#### HAProxy example config
+
+    # irrelevant parts omitted
```
juliusv: /cc @brian-brazil in case you want to take a look at these configs

README.md
```diff
@@ -88,6 +116,86 @@ To deploy PromDash behind a reverse proxy you can set a global path prefix
...
+        server main localhost:3000
+
+    backend prom
+        acl auth_ok http_auth(users_dash)
+        http-request auth unless auth_ok
+        server main localhost:9090
+
+#### Nginx example config
+
+    # irrelevant parts omitted
```
brian-brazil: Any example configuration we provide should be complete enough to work out of the box, without intimate knowledge of the given server to figure out the rest.

README.md
brian-brazil: I'm not sure these docs belong here, as this is more of a howto and quite a bit away from core functionality. Maybe a blog post?

README.md
juliusv: Ok yeah, makes sense. So I'd say either provide complete self-contained complete (and tested) configs here, or leave them out. That would also solve the problem of this README.md section getting quite long in general :)

README.md
juliusv: I agree, but have to focus my time on other things right now. I'd suggest adding it here now and optionally moving it later - or feel free to take over to make it a blog post :)

README.md
jonnenauha: I put "tabs" instead of "tab" because incognito wont re-show the prompt. I actually tested this now and it wont even promp if you use a new tab. You have to close all incognito windows then pop up a new one to test `/prom`. The problem is that if you first check `/` and it prompts, then `/prom` it will not prompt, this might be hiding the fact that the route in reverse proxy is not using auth.

Maybe I should remove this whole section and let for the user to figure out. Explaining all this will get kind of long. Not explaining it might leave you vulnerable if you screwed up the routing, though the fact we have the configs now should help out there.

I'll change it to "tab" as "tabs" did not have any real value as I expected.

README.md
jonnenauha: What about github wiki? If you enable it from the project settings I could put all this stuff there as well and have a link in the readme to it? Could then also have a link to it from the official docs or port it over there as a new dedicated page about reverse proxy setups.

README.md
jonnenauha: Yes, they do get very long at least for HAProxy that I use. Lot of boilerplate that is in the default config.

As said above, if it was a dedicated wiki page i could put longer ones. I will remove them for now from this pull request for now.

README.md
juliusv: I just meant "tab" because "a... tabs" (singular article for plural noun) wasn't grammatically correct :)

juliusv:
Sorry for being a bit slow on this PR. Hmm, we used the wiki once, but decided to close it because we already have two places for documentation: the README.mds and prometheus.io (which includes blog posts). I think long-term we will want a separate "Howtos" section on prometheus.io or something. 

For now, it's ok as it is now, after addressing https://github.com/prometheus/promdash/pull/449/files#r39730036 and https://github.com/prometheus/promdash/pull/449/files#r39730005.

Thanks!

juliusv:
(and then please squash your commits into a single one before merging)

jonnenauha:
Yeah, I'm in no rush here, just trying to help :)

I've used git for a very long time but have not had to use squash before. Lets see if i cant get this done without screwing things up!

jonnenauha:
Got a bit nervous about the `--force` push but looks like it did what you wanted. As this is a feature branch ofc force was fine, still :)

juliusv:
@jonnenauha Ah, better now, thanks :) What about this last comment from 6d ago? https://github.com/prometheus/promdash/pull/449/files#r39729955

jonnenauha:
I did not do that on my own setup. Will make it a little ugly and you have to remember to postfix your url with `/prom` when you want to go look at your metrics :P I guess one could do a redirect from the root to /prom in the reverse proxy by default. Seems kind of ugly but I see your point and people might make a prom named dashboard.

I'll change it.

juliusv:
@jonnenauha Thanks again for sticking with this :) Merging!

PR #454, opened 2015-09-18
Override Bootstrap's 0-padding for table cells.

juliusv:
The recent bootstrap upgrade "glued" together table cells when defining
annotation tags or dashboard variables by setting a 0-padding on td and
th.

Old:

bad

New:

good


juliusv:
@stuartnelson3 is away for some days, so merging for him to take a look later.

PR #456, opened 2015-09-23
correctly connect prometheus url and path in url generator

### LINKED ISSUE #450: Link to graph broken when using path in server URL
Noticed the `Datasources > Link to Graph in Prometheus` button link is broken if you use a path in the server url.

My server URL is `http://dash.example.com/prom/`
The link that button opens in a new tab is `http://dash.mapgets.com/promgraph#<fragment>`

It seems to strip the last fwd slash from the path + "graph". I dunno why without path it would work if it strips the first slash, that would become `http://dash.mapgets.comgraph#<fragment>` so there must be some special check for it.

stuartnelson3:
fixes #450 
@julius

### DIFF
--- a/app/assets/javascripts/angular/directives/expression.js.erb
+++ b/app/assets/javascripts/angular/directives/expression.js.erb
@@ -86,7 +86,7 @@ angular.module("Prometheus.directives").directive('expression', ["$http",
         }
 
         var expression = VariableInterpolator(scope.expr.expression, scope.vars);
-        var url = URLGenerator(server.url, "graph", scope.vars);
+        var url = URLGenerator(server.url, "/graph", scope.vars);
         var expr = {
           expr: expression,
           range_input: scope.range || "",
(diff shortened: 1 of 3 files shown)


juliusv:
:+1:

app/assets/javascripts/angular/directives/expression.js.erb
```diff
@@ -86,7 +86,7 @@ angular.module("Prometheus.directives").directive('expression', ["$http",
         }
 
         var expression = VariableInterpolator(scope.expr.expression, scope.vars);
-        var url = URLGenerator(server.url, "graph", scope.vars);
+        var url = URLGenerator(server.url, "/graph", scope.vars);
```
grobie: So, this change would not have been necessary actually?

stuartnelson3: Yes, totally unnecessary. It was added for consistency with the rest of the
project

Am Mittwoch, 23. September 2015 schrieb Tobias Schmidt :

PR #457, opened 2015-09-23
Upgrade capybara and selenium-webdriver to latest versions

samstarling:
This makes the tests pass on my local machine, because I was running a newer version of Firefox (40.0.2).

In past projects, I've had to keep these versions up to date fairly frequently. This is because Firefox auto-upgrades to a newer version, and breaks compatibility with the gems that drive the browser for tests.

/cc @juliusv @stuartnelson3

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,10 +19,10 @@ group :test, :development do
   gem 'sinatra'
   gem 'pry'
   gem 'rspec-rails', "~> 3.0.0.beta2"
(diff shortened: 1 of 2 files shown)


samstarling:
I'm going to wait and see what Travis CI says about this, and fix any problems as they arise.

samstarling:
The tests were far more successful this time :smile:

juliusv:
:+1: Cool, thanks :)

PR #461, opened 2015-10-08
fix errors in widget pages

stuartnelson3:
widget pages were failing to load annotations
because of a missing variable, which caused the
graph to become unresponsive

additionally, widget pages attempted to call a
function that did not exist.

@juliusv 

to test: create a widget page, verify no errors occur in console on load and hover detail exists. open a widget tab, hit "esc" to close, verify no error occurs in console.


juliusv:
:+1:

PR #463, opened 2015-11-11
Fix form markup on the 'Frame Source' popup

samstarling:
As pointed out by @stuartnelson3.


stuartnelson3:
looks good :+1: 

will merge once tests pass

PR #465, opened 2015-11-12
clear the annotation cache one switching profiles

stuartnelson3:
@samstarling 

hey, you got bandwidth to look at this today? i can swing by with an example to check out


samstarling:
Does this mean that when the context is changed, the annotations are cleared? I'm a little busy at the moment, but I'm free from 5pm on.

stuartnelson3:
yeah, so now as part of changing the profile, all annotations are cleared from the cache.

samstarling:
:+1:

PR #466, opened 2015-11-12
fix loading widgets that rely on profiles

stuartnelson3:
break out graphs will now load variables from the correct profile if one was selected

@samstarling check that this works for dashboards with and without profiles :)


samstarling:
:+1:

PR #468, opened 2015-11-19
Send credentials with XMLHttpRequests

amorken:
In cases where Prometheus instances run behind authenticating proxies,
cookies or Basic authentication may be required for requests from
PromDash too.

I'm no web dev, this change was devised purely by parroting the Angular docs.
Seems to work on our deployment though.

@stuartnelson3 , you probably have a clue, though, and I'd appreciate it if you could take a look. Thanks. =)

### DIFF
--- a/app/assets/javascripts/angular/app.js
+++ b/app/assets/javascripts/angular/app.js
@@ -19,6 +19,7 @@ angular.module("Prometheus",
   if ($location.hash()[0] === "?") {
     $location.url($location.path() + $location.hash());
   }
-}]).config(['$locationProvider', function($locationProvider) {
+}]).config(['$locationProvider', '$httpProvider', function($locationProvider, $httpProvider) {
   $locationProvider.html5Mode(true);
+  $httpProvider.defaults.withCredentials = true;
 }]);


stuartnelson3:
hey, when testing this locally I get an error:

```
XMLHttpRequest cannot load http://prometheus-server:9090/api/metrics. A wildcard '*' cannot be used in the 'Access-Control-Allow-Origin' header when the credentials flag is true.
```

this change breaks requests sent directly to a prometheus instance. (because you're making your request to an authenticating proxy which then forwards the request, everything works fine for you).

by default, prometheus sets its `Access-Control-Allow-Origin` as `*`, which (as you can see above) is a no go when also sending credentials.

because it is not currently possible to configure this header in prometheus, this would be breaking behavior for everyone connecting directly.

i've chatted with @fabxc and @beorn7 and let them know about the situation. you'll have to talk with them about adding this.

additionally, if this were added, all users would be forced to configure their prometheus servers to use the correct header in order for their promdash to pull the data.

i'm going to close this pr for now pending things being resolved on the prometheus end.

amorken:
Yes, that's true, I forgot about Prometheus' hardcoded ACAO header as we hack that up in our nginx proxy anyway. But I'll see about throwing together a patch for Prometheus proper as well. =)

Thanks,
-@M

PR #475, opened 2016-01-14
[FEATURE] User written functions for getting line color

samdroid-apps:
Allow users to write functions that return the color of the line.  By
default, the function just returns the next color from the palette
(same as current behaviour).  Hoewver, the user can use the labels and
color palette object to find a new color.  For example:

    (function (labels, palette) {
      if (labels.bad === 'True') {
        return 'red';
      } else {
        return palette.color();
      }
    })

They can even use d3, which is one of the globals:

    (function (labels, palette) {
      if (labels.bad === 'True') {
        return d3.rgb(palette.color()).darken();
      } else {
        return palette.color();
      }
    })

### DIFF
--- a/app/assets/javascripts/angular/controllers/graph_ctrl.js
+++ b/app/assets/javascripts/angular/controllers/graph_ctrl.js
@@ -1,3 +1,5 @@
+var DEFAULT_COLOR_FUNC = '(function (labels, palette) { return palette.color(); })';
+
 angular.module("Prometheus.controllers").controller('GraphCtrl',
                                                     ["$scope",
                                                       "$http",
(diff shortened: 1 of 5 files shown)


samdroid-apps:
Sorry, I didn't add a jshint exception for the eval.  I have amended to commit, and hopefully the tests will be green this time.

stuartnelson3:
Hey, thanks for the submission! I'll look at this ASAP, catching up after vacation

samdroid-apps:
Thanks.

I just force pushed a very small change, which bound colorFunc change to redraw the graph, similar to the other options.

PR #477, opened 2016-01-15
Display label in hoverDetail if label is from format string

samdroid-apps:
This allows for human friendly labels to be shown in charts with
many series (so showing the legend would take too much space).

Similar to #368, but with a more limited scope

E.g. with custom format string "test!":
[https://gyazo.com/0b89ff00a2eb3ed983980bb8d6ccd4c6](https://gyazo.com/0b89ff00a2eb3ed983980bb8d6ccd4c6)

E.g. without a format string:
[https://gyazo.com/aa63ccceb29b2206578605dd506c9f03](https://gyazo.com/aa63ccceb29b2206578605dd506c9f03)

PR #478, opened 2016-01-17
Highlight the background of future graph sections

samdroid-apps:
This is a simple usability aid that allows users to better orient
themself in time.  This is especially useful when users are not
familiar with UTC times.

Eg.  dark graph:

[https://gyazo.com/b58bf5dc1f102d35691367296017768b](https://gyazo.com/b58bf5dc1f102d35691367296017768b)

Eg.  light graph

[https://gyazo.com/c69a04da4976cf8e55eaff3afa12a68a](https://gyazo.com/c69a04da4976cf8e55eaff3afa12a68a)

### DIFF
--- a/app/assets/javascripts/angular/directives/graph_chart.js
+++ b/app/assets/javascripts/angular/directives/graph_chart.js
@@ -311,6 +311,9 @@ angular.module("Prometheus.directives").directive('graphChart', [
           height: calculateGraphHeight($legend)
         });
         rsGraph.series.legend = legend;
+        rsGraph.onUpdate(function () {
+          highlightFuture(rsGraph);
+        });
         rsGraph.render();
 
         AnnotationRefresher.refresh(scope.graphSettings.tags, scope.graphSettings.range, scope.graphSettings.endTime, scope.vars)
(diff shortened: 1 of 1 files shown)


samdroid-apps:
I just found a breaking case unfortunately.  Steps to reproduce:

1.  Have graph
2.  See future highlighted
3.  Hide one of the lines (using the legend)
4.  Notice that the future highlight is now gone

This can be very easily fixed however, and I have force pushed a change to do so.

samdroid-apps:
Hum, I'm not 100% sure about the test failure:

```
Failures:

  1) Graph legend general behavior maintains legend state after update
     Failure/Error: legend_elements.first.click
     NoMethodError:
       undefined method `click' for nil:NilClass
     # ./spec/features/graph/legend_spec.rb:23:in `block (3 levels) in <top (required)>'
```

Especially since this change has nothing to do with legend elements.  This was also the test failure on #477, even though that had nothing to do with legend element either.

stuartnelson3:
that's one of the random test failures that unfortunately we suffer from

stuartnelson3:
I'm not sure that this visual aid is really necessary, and it's my hope that users will become accustomed to UTC time

beorn7:
What's the use-case for displaying the future at all?
Prometheus cannot cleanly display metrics from the future ever. So why would you ever display a time range that is in the future with PromDash? "Now" is always the right edge, unless you are deliberately displaying historical data.
(I could see this a valid use-case with other data sources, but PromDash doesn't aspire to be a generic dashboarding tool for arbitrary data sources. Use Grafana if you want that.)

samdroid-apps:
I find it very confusing navigating without a visual aid of where now is.  Eg.  if I go back in the 6h zoom level, zoom out to 12h and come forward in time I get into a situation where now is in the middle of the graph.

I am actually using PromDash/Prometheus to display data in the future, and this feature helps navigation.

matthiasr:
This seems like the actual problem to me – let's fix this instead of visual
workarounds?

beorn7:
That would be my gut feeling, too: never display the future in Promdash.

PR #482, opened 2016-03-07
Add RAILS_RELATIVE_URL_ROOT to proxy documentation

### LINKED ISSUE #434: Setting PROMDASH_PATH_PREFIX does not work for assets

turb:
Fixes #434.

- `asset_path` method does not use `PROMDASH_PATH_PREFIX` but takes `RAILS_RELATIVE_URL_ROOT`
- `routes.rb` ignores `RAILS_RELATIVE_URL_ROOT` but explictely uses `PROMDASH_PATH_PREFIX`

Therefore, the workaround is to add both configurations. The best, and more invasive fix will be to refactor to take care of only one env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,8 @@ You can pass parameters directly to the thin server with:
 ### Deploy behind a reverse proxy
 
 To deploy PromDash behind a reverse proxy you can set a global path prefix
-using the environment variable `PROMDASH_PATH_PREFIX`. Once set all URLs will
-start with the given prefix.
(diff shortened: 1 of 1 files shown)


grobie:
Thanks!

PR #484, opened 2016-04-27
Update API calls

stuartnelson3:
@juliusv @fabxc can you guys double check the code, and run against a local instance? I checked against pie/gauge/graph, but it wasn't exhaustive.

addresses https://github.com/prometheus/prometheus/issues/1593

TODO:

- Update test fixtures


app/assets/javascripts/angular/services/graph_refresher.js
```diff
@@ -47,12 +47,12 @@ angular.module("Prometheus.services").factory('GraphRefresher',
...
         return deferred.promise;
       },
-      graphite: function(idx, expression, server, expressionID, endTime, range, step) {
+      graphite: function(idx, expression, server, expressionID, startTime, endTime, step) {
         var deferred = $q.defer();
         $http.get(URLGenerator(server.url, 'render', $scope.vars), {
           params: {
             target: expression,
-            from: GraphiteTimeConverter.graphiteFrom(range, endTime),
+            from: GraphiteTimeConverter.graphiteUntil(startTime),
```
juliusv: Why is this `graphiteUntil` now instead of `graphiteFrom`?

juliusv: Ah, I guess I get it - but is the `graphiteUntil` function still named well then?

juliusv:
I tested everything I could think of (different widget types, errors, successes, different time ranges), and the only error I've found so far is when you try to change the end time. Then you get:

`endTime.getTime is not a function`

app/assets/javascripts/angular/services/graph_refresher.js
stuartnelson3: Nope, the name is totally misleading!

stuartnelson3:
@juliusv thanks for checking, like the old days :) anything else you want changed?

juliusv:
Nope! :+1: otherwise. I'm not gonna go into nitpicking mode for basically deprecated code :)

stuartnelson3:
Haha yeah, for definitely deprecated code :)

fabxc:
\o/ great thanks!

PR #486, opened 2016-10-07
Restrict log scale domain min value to 1 to prevent NaN.

turb:
Log scale may be very convenient. However, we happen to have some values at 0.This should not be a problem since this is a log scale (ie just a way to display).

However, if we have a 0 value, the complete series is erased, replaced by "NaN". The trivial workaround is to put "+1" on the query (urk).

This commit reduces the scale domain to [1, max] and prevents this bug. The series is now considered as "no value" only on dots where the value is 0 (interrupted curve) which is better than nothing.

### DIFF
--- a/app/assets/javascripts/angular/services/y_axis_utilities.js
+++ b/app/assets/javascripts/angular/services/y_axis_utilities.js
@@ -13,7 +13,7 @@ angular.module("Prometheus.services").factory('YAxisUtilities', [function() {
   }
   return {
     setLogScale: function(min, max) {
-      logScale = extendScale(d3.scale.log().domain([min, max]));
+      logScale = extendScale(d3.scale.log().domain([1, max]));
       return logScale;
     },
     setLinearScale: function(min, max) {


stuartnelson3:
Hey,

PromDash is officially deprecated, and has been for quite a while. We recommend using grafana, as it has native prometheus support, and a fuller feature set.

turb:
@stuartnelson3 wow, didn't see that, as Promdash works pretty well for us. We should look for a migration, then.

However, merging this should not hurt anybody ;)

stuartnelson3:
Yeah, PromDash is pretty awesome.

What's the complete range of the data you're using that causes values to be `NaN`? I remember when this was originally being implemented that I bounded values to 0 so that they would render properly, so I suppose I'm curious why for your team this isn't working.

Relevant code: https://github.com/prometheus/promdash/blob/master/app/assets/javascripts/angular/directives/graph_chart.js#L109-L123

turb:
The answer is easy: we have some 0 values, and log(0)=NaN ;)

Another correct fix should be then to amend your code and replace the 0 minimum to a 1 (if between 0 and 1, the log would be negative). However, for what I understand, the D3js `domain()` function already provides this.

stuartnelson3:
Hm, good point.

I suppose my only concern is for people (if there even are any) that are using a log scale but looking at a data range that exists between 0 and 1. Seems pretty unlikely, though.

I'll go ahead and merge this. We really do recommend migrating to grafana as soon as possible, as it is a superior solution.

stuartnelson3:
Apparently, actually, I have had my merge permissions revoked -.-

